PR #12, opened 2024-01-30
update(billing): update instructions for assigning projects to billing groups

staceysalamon-aiven:
## Describe your changes
Update instructions for assigning projects to billing groups to reflect changes in the UI.

## Checklist

- [X] I checked the output.

### DIFF
--- a/docs/platform/howto/billing-assign-projects.md
+++ b/docs/platform/howto/billing-assign-projects.md
@@ -2,26 +2,15 @@
 title: Assign projects to billing groups
 ---
 
-When you create a new project, you can assign it to a billing group,
-which lets you
-[use the same payment card and consolidate costs](/docs/platform/concepts/billing-groups). You can also assign the project to another billing group
-later in the [Aiven Console](https://console.aiven.io/) using the
-instructions in this article.
+You can assign multiple projects to a billing group to [use the same payment card and consolidate the costs](/docs/platform/concepts/billing-groups).
 
-:::note
-Projects can only be assigned to one billing group. If a project is
-already assigned to one billing group and you assign it to another, it
-will be removed from the first group.
-:::
+Each project can only be assigned to one billing group. If a project is already assigned
+to a billing group and you assign it to another, it will be removed from the first group.
 
 To assign projects to a billing group:
 
-1.  In the project, click **Billing**.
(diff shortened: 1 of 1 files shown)


docs/platform/howto/billing-assign-projects.md
```diff
@@ -2,26 +2,15 @@
...
-5.  Select the projects to assign to this billing group and click
-    **Assign projects**.
-6.  Click **Cancel** to close the dialog box.
-
-You will see the newly assigned projects listed in this billing group.
+1.  Go to **Billing** > **Billing groups**.
+1.  Select the billing group you want to assign the project to.
+1.  On the **Projects** tab, click **Assign projects**.
+1.  Select the projects and click **Assign projects**.
+1.  Click **Cancel** to close the dialog box.
```
ArthurFlageul, line 16: This doesn't sound like a possible step after clicking **Assign projects**, but it looks like an alternative to this step. I'd recommend you either insert this instruction in the step on line 15 as an alternative final step, or delete this step altogether.

In the spirit of minimalism, I would highly recommend we never document obvious cancelation steps.

ArthurFlageul (changes requested):

docs/platform/howto/billing-assign-projects.md
staceysalamon-aiven: You can test this in the Console and see that that is how it works. This will likely change with future changes to the billing section because it's obviously not a great UX, but for now that's the flow. The reason for adding the step is that I don't think it's obvious after you assign the projects that they are in fact now assigned to the BG (that's how it felt to me when I tested it). You really have to close the dialog to see this. I agree adding obvious steps is unnecessary. I can remove this here; I just think it's a special case where maybe reassuring anyone reading that they haven't missed something makes sense.

ArthurFlageul: I haven't tested the UI but if to terminate the workflow, we indeed have to click `Cancel`, then we're good for this PR.
On the other hand that's a text-book example of what I mean by "imagining the docs of the UI helps us fix the UI". Seeing ourselves finishing a task by "click cancel" is obviously a red flag about the UI that we should absolutely catch in the UI design phase.

ArthurFlageul (approved):
Nice content reduction 👍

PR #20, opened 2024-01-31
feat: required access snippet

staceysalamon-aiven:
## Describe your changes
Adds a snippet for inserting a section into docs indicating what permissions are needed to access a feature.

### DIFF
--- a/.vscode/aiven.code-snippets
+++ b/.vscode/aiven.code-snippets
@@ -21,7 +21,7 @@
       "",
       "You must be a ${1|[super admin](/docs/platform/howto/make-super-admin),[project admin](/docs/platform/reference/project-member-privileges),[project operator](/docs/platform/reference/project-member-privileges)|} to access this feature."
     ],
-    "description": "Inserts a section with information on what role is needed for a feature"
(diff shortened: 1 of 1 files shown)


.vscode/aiven.code-snippets
```diff
@@ -13,6 +13,17 @@
     "description": "Inserts a related pages section"
   },
 
+  "Required access": {
+    "scope": "markdown",
+    "prefix": "Required access",
+    "body": [
+      "## Required access",
```
ArthurFlageul, line 20: Nice to see you're looking into snippets 👌  

As far as the content goes, we should do our best not to duplicate titles, so we don't end up with many identical H2 by design and negatively impact SEO. Maybe this should be an admonition?

ArthurFlageul (changes requested):

.vscode/aiven.code-snippets
ArthurFlageul: I also realize we certainly did this for the related links bit but I'll modify it to not make it a _real_ h2.

staceysalamon-aiven:
@ArthurFlageul Made the changes, but Cloudflare is now complaining about a bunch of broken links. 🤔

PR #27, opened 2024-02-01
update(platform): trial limitations

staceysalamon-aiven:
## Describe your changes

Update limitations as trial credits will now be assigned at org level. Other minor tweaks to links, line length, and paring back language.

## Checklist

- [X] I checked the output.


docs/platform/concepts/free-trial.md
```diff
@@ -2,43 +2,30 @@
...
--   Trials can only be started in your default project, which is created
-    when you sign up
--   Trial credits cannot be transferred to another project
--   You can\'t create new services if your remaining credits would be
-    spent too quickly
+There are some limitations:
 
-## Add a payment card
+-   You can only have one trial.
+-   Trial credits are added to the default
```
vtainio, line 25: This is something that has changed actually a bit but I'm not sure how important the changes are. Technically we add the trial credits to the billing group where the user performs the trial starting action (like creating a paid service). This isn't necessarily the default billing group but I would assume that in 95% of the trials it will be because people aren't creating billing groups for fun.

So not sure if this needs to be changed at all but we could just maybe say something like "Trial credits are assigned to a single billing group".

staceysalamon-aiven: Good to know, thanks! I've updated that part as suggested.

ArthurFlag (approved):

PR #51, opened 2024-02-13
update: CLI instructions with Homebrew installation instructions

staceysalamon-aiven:
## Describe your changes

Add instructions for installing Aiven client via Homebrew.

## Checklist

- [X] I checked the output.

### DIFF
--- a/docs/tools/cli.md
+++ b/docs/tools/cli.md
@@ -4,11 +4,6 @@ title: Aiven CLI
 
 The Aiven command line interface (CLI) lets you use the Aiven platform and services in a scriptable way through the API.
 
-It's available [on GitHub](https://github.com/aiven/aiven-client). Check out
-[this blog](https://aiven.io/blog/aiven-cmdline) to learn how to use the Aiven CLI
-for common tasks. If you prefer video, watch the short
-[how to get started tutorial](https://www.youtube.com/watch?v=nf3PPn5w6K8).
-
 ## Get started
 
 ### Install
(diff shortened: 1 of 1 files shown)


docs/tools/cli.md
```diff
@@ -2,57 +2,54 @@
...
-[this blog](https://aiven.io/blog/aiven-cmdline) to learn how to use the
-Aiven command line tool to do common tasks.
+The Aiven command line interface (CLI) lets you use the Aiven platform and services in a scriptable way through the API.
 
-If you prefer to follow a video tutorial, check out this short video on
-[how to get started](https://www.youtube.com/watch?v=nf3PPn5w6K8).
+It's available [on GitHub](https://github.com/aiven/aiven-client). Check out
+[this blog](https://aiven.io/blog/aiven-cmdline) to learn how to use the Aiven CLI
+for common tasks. If you prefer video, watch the short
+[how to get started tutorial](https://www.youtube.com/watch?v=nf3PPn5w6K8).
```
ArthurFlag, line 10: Small suggestions:

- Could we avoid patterns "this blog" "this page" and "here" for links?
It's more efficient and predictive to use a verb that's usually the title of the page. In this case something like `[Learn more about the Aiven CLI](link) or see the [How to get started](link) video.`
- Let's avoid describing things, like "short video" or "easy installation", "quick way to do x".  Maybe 5 minutes is too long for some our readers :) 

I'm also unsure about redirecting ppl to a blog post from the top of the page instead of teaching people things on this page directly. There's a bit of repeat between the 2 pages, for example, the installation method.  Could we think of adding this as a related page link perhaps, wdyt?

ArthurFlag (changes requested):

docs/tools/cli.md
staceysalamon-aiven: Yes, agreed, but I've opted for fixing the absolutely necessary changes flagged by Vale to save time and then really just adding the Homebrew instructions. There's a lot that needs to be done both on this page and the CLI section in general, but I wanted to get the Homebrew instructions added here as one of the teams pointed out we didn't mention it anywhere in the docs.

ArthurFlag (approved):

PR #59, opened 2024-02-15
fix: Fixed the guide to monitor metrics on Grafana prebuilt dashboards

palmieric:
The guides to monitor Kafka and PostgreSQL using Aiven prebuilt dashboards, wrongly referenced to create an integration of type Monitor Data in Grafana instead of Grafana Metrics Dashboard.

Fixed it for both service types.


wojcik-dorota (approved):
LGTM - thank you!

PR #60, opened 2024-02-15
add(mysql postgresql): high availability information

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-675

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/mysql/concepts/high-availability-mysql.md
+++ b/docs/products/mysql/concepts/high-availability-mysql.md
@@ -93,8 +93,8 @@ node fails while all the standby nodes are being recovered:
 
 - On a two-standby node plan
 
-  To avoid a cascading failure condition leading to data lost, the second failover to the
-  last standby is recommended only after a new standby is provisioned.
+  The second failover to the last standby is recommended only after a new standby is
+  provisioned.
 
 :::note
 The amount of time it takes to replace a failed node depends mainly on
--- a/docs/products/mysql/concepts/high-availability-mysql.md
+++ b/docs/products/mysql/concepts/high-availability-mysql.md
@@ -3,7 +3,7 @@ title: High availability of Aiven for MySQL®
 sidebar_label: High availability
 ---
 
-Aiven for MySQL® is available on a variety of plans, offering different levels of high availability. The selected plan defines the features available, and a summary is provided in the table below:
+Aiven for MySQL® is available on a variety of plans, offering different levels of high availability. The selected plan defines the features available.
 
 | Plan         | High availability features                                      | Backup history | Uptime  |
 | ------------ | --------------------------------------------------------------- | -------------- | ------- |
@@ -33,7 +33,8 @@ Minor failures, such as service process crashes or temporary loss of
 network access, are handled automatically by Aiven in all plans without
 any major changes to the service deployment. The service automatically
 resumes normal operation once the crashed process is automatically restarted or when
-network access is restored.
+network access is restored. On a plan with a standby node, a failover is performed if a
+crash loop occurs.
 
 ### Severe failures
 
@@ -55,9 +56,9 @@ primary node.
 
 - With standby, recovery time objective (RTO) is quick, and recovery point objective (RPO)
   is close to zero loss.
-- Without standby, RPO loses 5 minutes of data, and RTO is variable as it involves
+- Without standby, RPO loses up to 5 minutes of data, and RTO is variable as it involves
   restoring a backup and applying binlogs, which depends on the database size and the
-  write load (up to 6 hours or more in edge cases).
+  write load (up to several hours or more in edge cases).
 
 ## Highly available Business and Premium service plans
 
@@ -86,10 +87,10 @@ node fails while all the standby nodes are being recovered:
 - On a single-standby node plan
 
   New nodes are automatically scheduled for creation to become the new primary and standby.
-  The primary node is restored from the latest available backup, which can involve some
-  degree of data loss. Any write operations made since the backup of the binary log file
-  are lost. Typically, this time window is limited to either five minutes of time or one
-  binary log file.
+  The primary node is restored from the latest available backup. Next, writes made since
+  the last backup are applied from the backup of the binary log file. Any writes made
+  since the backup of the binary log file are lost. Typically, this time window is limited
+  to either five minutes of time or one binary log file.
 
 - On a two-standby node plan
 
@@ -104,7 +105,7 @@ keeps on serving clients even during the recreation of the other node.
 All of this is automatic and requires no administrator intervention.
 :::
 
-**Premium** plans operate in a similar way as **Business** plans. The
+Premium plans operate in a similar way as Business plans. The
 main difference comes when one of the standby nodes or the primary node
 fails. Premium plans have an additional, redundant standby node
 available, providing platform availability even in the event of losing
@@ -127,5 +128,5 @@ available backup also applying the saved binlogs. Next, it resumes serving custo
 
 Since there is just a single node providing the service, the service is
 unavailable for the duration of the restoration. In addition, any write
-operations made since the backup of the latest binary log file are lost. Typically, this time
-window is limited to either five minutes of time or one binary log file.
+operations made since the backup of the latest binary log file are lost. Typically, this
+time window is limited to either five minutes of time or one binary log file.
--- a/docs/products/mysql/concepts/high-availability-mysql.md
+++ b/docs/products/mysql/concepts/high-availability-mysql.md
@@ -3,16 +3,14 @@ title: High availability of Aiven for MySQL®
 sidebar_label: High availability
 ---
 
-Aiven for MySQL® is available on a variety of plans, offering
-different levels of high availability. The selected plan defines the
-features available, and a summary is provided in the table below:
+Aiven for MySQL® is available on a variety of plans, offering different levels of high availability. The selected plan defines the features available, and a summary is provided in the table below:
 
- | Plan         | High Availability Features                                                     | Backup History |
- | ------------ | ------------------------------------------------------------------------------ | -------------- |
- | **Hobbyist** | Single-node with limited availability                                          | 2 days         |
- | **Startup**  | Single-node with limited availability                                          | 2 days         |
- | **Business** | Two-node (primary + standby) with higher availability                          | 14 days        |
- | **Premium**  | Three-node (primary + standby + standby) with top availability characteristics | 30-day         |
+| Plan         | High availability features                                      | Backup history | Uptime  |
+| ------------ | --------------------------------------------------------------- | -------------- | ------- |
+| **Hobbyist** | Single node with limited availability                           | 2 days         | 98.5 %  |
+| **Startup**  | Single node with limited availability                           | 2 days         | 99 %    |
+| **Business** | Two nodes (primary + standby: basic high availability)          | 14 days        | 99.9 %  |
+| **Premium**  | Three nodes (primary + 2 x standby: stronger high availability) | 30-day         | 99.99 % |
 
 ## About primary and standby nodes
 
@@ -30,14 +28,18 @@ service is useful for multiple reasons:
 
 ## Failure handling
 
-**Minor failures**, such as service process crashes or temporary loss of
+### Minor failures
+
+Minor failures, such as service process crashes or temporary loss of
 network access, are handled automatically by Aiven in all plans without
 any major changes to the service deployment. The service automatically
-restores normal operation once the crashed process is automatically
-restarted or when network access is restored.
+restores normal operation once the crashed process is restarted or when network access is
+restored.
 
-**Severe failures**, such as losing a node entirely in case of hardware
-or severe software problems, require drastic recovery measures. The
+### Severe failures
+
+Severe failures, such as losing a node entirely in case of hardware
+or severe software problems, require involved recovery measures. The
 Aiven monitoring infrastructure automatically detects a failing node
 both when the node starts reporting issues in the self-diagnostics or
 when stops communicating. In such cases, the monitoring infrastructure
@@ -51,25 +53,28 @@ primary node.
 
 ## Highly available Business and Premium service plans
 
-When the failed node is a MySQL **standby** node, the primary node
+When the failed node is a MySQL standby node, the primary node
 keeps running normally and provides a normal service level to the client
 applications. Once the new replacement standby node is ready and
 synchronised with the primary node, it starts replicating the primary
 node in real time as the situation reverts back to normal.
 
-When the failed node is a MySQL **primary** node, the combined
+When the failed node is a MySQL primary node, the combined
 information from the Aiven monitoring infrastructure and the standby
 node is used to make a failover decision. The standby node is then
 promoted as the new primary and immediately starts serving clients. A
 new replacement node is automatically scheduled and becomes the new
-standby node.
+standby node. There is no data loss in this scenario.
 
-If all the **primary** and **standby nodes** fail at the same time, new
+If all the primary and standby nodes fail at the same time or if all the primary
+nodes fail while standby nodes are being recovered, new
 nodes are automatically scheduled for creation to become the new primary
 and standby. The primary node is restored from the latest available
-backup, which could involve some degree of data loss. Any write
+backup, which can involve some degree of data loss. Any write
 operations made since the backup of the latest write-ahead logging (WAL) file are lost.
 Typically, this time window is limited to either five minutes of time or one WAL file.
+
+If all the primary nodes fail while a standby is being recovered
 
 :::note
 The amount of time it takes to replace a failed node depends mainly on
@@ -80,7 +85,7 @@ All of this is automatic and requires no administrator intervention.
 :::
 
 **Premium** plans operate in a similar way as **Business** plans. The
-main difference comes when one of the **standby** or **primary** nodes
+main difference comes when one of the standby or primary nodes
 fails. Premium plans have an additional, redundant standby node
 available, providing platform availability even in the event of losing
 two nodes. In cases where the primary node fails, Aiven monitoring tool determines which
--- a/docs/products/mysql/concepts/high-availability-mysql.md
+++ b/docs/products/mysql/concepts/high-availability-mysql.md
@@ -1,0 +1,107 @@
+---
+title: High availability of Aiven for MySQL®
+sidebar_label: High availability
+---
+
+Aiven for MySQL® is available on a variety of plans, offering
+different levels of high availability. The selected plan defines the
+features available, and a summary is provided in the table below:
+
+ | Plan         | High Availability Features                                                     | Backup History |
+ | ------------ | ------------------------------------------------------------------------------ | -------------- |
+ | **Hobbyist** | Single-node with limited availability                                          | 2 days         |
+ | **Startup**  | Single-node with limited availability                                          | 2 days         |
+ | **Business** | Two-node (primary + standby) with higher availability                          | 14 days        |
+ | **Premium**  | Three-node (primary + standby + standby) with top availability characteristics | 30-day         |
+
+## About primary and standby nodes
+
+Aiven's Business and Premium plans offer primary nodes and standby nodes. A standby
+service is useful for multiple reasons:
+
+-   Provides another physical copy of the data in case of hardware,
+    software, or network failures
+-   Typically reduces the data loss window in disaster scenarios
+-   Provides a quicker database time to restore with a controlled
+    failover in case of failures, as the standby is already installed,
+    running, and synchronised with the data
+-   Can be used for read-only queries to reduce the load on the primary
+    server
+
+## Failure handling
+
+**Minor failures**, such as service process crashes or temporary loss of
+network access, are handled automatically by Aiven in all plans without
+any major changes to the service deployment. The service automatically
+restores normal operation once the crashed process is automatically
+restarted or when network access is restored.
+
+**Severe failures**, such as losing a node entirely in case of hardware
+or severe software problems, require drastic recovery measures. The
+Aiven monitoring infrastructure automatically detects a failing node
+both when the node starts reporting issues in the self-diagnostics or
+when stops communicating. In such cases, the monitoring infrastructure
+automatically schedules a new replacement node to be created.
+
+:::note
+In the event of database failover, the **Service URI** of your service
+remains the same; only the IP address will change to point to the new
+primary node.
+:::
+
+## Highly available Business and Premium service plans
+
+When the failed node is a MySQL **standby** node, the primary node
+keeps running normally and provides a normal service level to the client
+applications. Once the new replacement standby node is ready and
+synchronised with the primary node, it starts replicating the primary
+node in real time as the situation reverts back to normal.
+
+When the failed node is a MySQL **primary** node, the combined
+information from the Aiven monitoring infrastructure and the standby
+node is used to make a failover decision. The standby node is then
+promoted as the new primary and immediately starts serving clients. A
+new replacement node is automatically scheduled and becomes the new
+standby node.
+
+If all the **primary** and **standby nodes** fail at the same time, new
+nodes are automatically scheduled for creation to become the new primary
+and standby. The primary node is restored from the latest available
+backup, which could involve some degree of data loss. Any write
+operations made since the backup of the latest write-ahead logging (WAL) file are lost.
+Typically, this time window is limited to either five minutes of time or one WAL file.
+
+:::note
+The amount of time it takes to replace a failed node depends mainly on
+the selected cloud region and the amount of data to be restored.
+However, in the case of partial loss of the cluster, the surviving node
+keeps on serving clients even during the recreation of the other node.
+All of this is automatic and requires no administrator intervention.
+:::
+
+**Premium** plans operate in a similar way as **Business** plans. The
+main difference comes when one of the **standby** or **primary** nodes
+fails. Premium plans have an additional, redundant standby node
+available, providing platform availability even in the event of losing
+two nodes. In cases where the primary node fails, Aiven monitoring tool determines which
+of the standby nodes is the furthest along
+in replication (has the least potential for data loss) and does a
+controlled failover to that node.
+
+:::note
+For Aiven for MySQL backups and restoration, Aiven uses
+[MyHoard](https://aiven.io/blog/introducing-myhoard-your-single-solution-to-mysql-backups-and-restoration).
+:::
+
+## Single-node Hobbyist and Startup service plans
+
+Hobbyist and Startup plans provide a single node; when it's lost, Aiven
+immediately starts the automatic process of creating a new replacement
+node. The new node starts up, restores its state from the latest
+available backup, and resumes serving customers.
+
+Since there is just a single node providing the service, the service is
+unavailable for the duration of the restoration. In addition, any write
+operations made since the backup of the latest WAL file are lost. Typically, this time
+window is limited to
+either five minutes of time or one WAL file.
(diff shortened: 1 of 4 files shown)


docs/products/postgresql/concepts/high-availability.md
```diff
@@ -30,14 +30,18 @@ Aiven's Business and Premium plans offer
 
 ## Failure handling
 
-**Minor failures**, such as service process crashes or temporary loss of
+### Minor failures
+
+Minor failures, such as service process crashes or temporary loss of
 network access, are handled automatically by Aiven in all plans without
 any major changes to the service deployment. The service automatically
-restores normal operation once the crashed process is automatically
```
divdby0, line 36: I think we should keep the automatic word here. Automatic fail-overs is a differentiator feature. There is also manual fail-overs.

docs/products/postgresql/concepts/high-availability.md
```diff
@@ -30,14 +30,18 @@ Aiven's Business and Premium plans offer
...
-restarted or when network access is restored.
+restores normal operation once the crashed process is restarted or when network access is
+restored.
+
+### Severe failures
 
-**Severe failures**, such as losing a node entirely in case of hardware
-or severe software problems, require drastic recovery measures. The
+Severe failures, such as losing a node entirely in case of hardware
+or severe software problems, require involved recovery measures. The
```
divdby0, line 44: The previous drastic word might sound a little bit "drastic" :) but I don't fully get the involved word in exchange. Both minor and severe failures are automatically handled by the system and they don't require any involvement from the user. The primary difference is during the severe failures, the system might take more serious measures to replace a node, on behalf of the user -- hence the "drastic" part.

docs/products/postgresql/concepts/high-availability.md
```diff
@@ -51,26 +55,27 @@ primary node.
...
 node in real time as the situation reverts back to normal.
 
-When the failed node is a PostgreSQL **primary** node, the combined
+When the failed node is a PostgreSQL primary node, the combined
 information from the Aiven monitoring infrastructure and the standby
 node is used to make a failover decision. The standby node is then
 promoted as the new primary and immediately starts serving clients. A
 new replacement node is automatically scheduled and becomes the new
-standby node.
+standby node. There is no data loss in this scenario.
```
divdby0, line 69: I think there is always a risk for loss of data. There might be edge cases where the standby is behind the master or the promotion of the standby might take longer. I think we should soften this or find another perspective to create contrast between this scenario and the previous one (standby fail).

docs/products/postgresql/concepts/high-availability.md
```diff
@@ -51,26 +55,27 @@ primary node.
...
 
-If all the **primary** and **standby nodes** fail at the same time, new
+If all the primary and standby nodes fail at the same time or if all the primary
```
divdby0, line 71: There is only one primary node at any given time. I am not sure I follow this part. I suggest to rever to the previous version.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+| **Premium**  | Three nodes (primary + 2 x standby: stronger high availability) | 30-day         | 99.99 % |
+
+## About primary and standby nodes
+
+Aiven's Business and Premium plans offer primary nodes and standby nodes. A standby
+service is useful for multiple reasons:
+
+-   Provides another physical copy of the data in case of hardware,
+    software, or network failures
+-   Typically reduces the data loss window in disaster scenarios
```
jfg-aiven, line 22: I feel this sentence is overly complex (typically and introduction of the notion of disaster).  I would simplify to "Reduces the data loss window in failure scenarios".

docs/products/postgresql/concepts/high-availability.md
```diff
@@ -1,23 +1,23 @@
...
+| ------------ | --------------------------------------------------------------- | -------------- | ------- |
+| **Hobbyist** | Single node with limited availability                           | 2 days         | 98.5 %  |
+| **Startup**  | Single node with limited availability                           | 2 days         | 99 %    |
+| **Business** | Two nodes (primary + standby: basic high availability)          | 14 days        | 99.9 %  |
+| **Premium**  | Three nodes (primary + 2 x standby: stronger high availability) | 30-day         | 99.99 % |
 
 ## About primary and standby nodes
 
 Aiven's Business and Premium plans offer
 [primary nodes](/docs/products/postgresql/reference/terminology) and
```
divdby0, line 18: I think the confusion around primaries comes from this original line. There is always one primary and 1/2 standby(s) in our plans. I think the original wording was confusing already as it reads like "multiple primaries" in a cluster which is incorrect. Otherwise it would be a multi-master/multi-primary solution.

I think the correct one-liner expression should be "Aiven's Business and Premium plans offer a primary node and standby nodes." Even this would still be ambigious because Business plans have only one standby node whereas Premium plans have 2.

Please also refer to my other comment for "primaries" below.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+-   Provides a quicker database time to restore with a controlled
```
jfg-aiven, line 23: I am not sure it is a quicker time to restore, as availability in these plans do not depend on a restore, but on the presence of a standby node.  I would rewrite as "Provides higher availability as failure recovery leverages already provisioned standby node(s) and does not suffer the delay of restoring a backups".

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+    failover in case of failures, as the standby is already installed,
+    running, and synchronised with the data
+-   Can be used for read-only queries to reduce the load on the primary
```
jfg-aiven, line 26: Care with this: sending reads on a standby node might be appealing, but might not work well when a failure happens:
- when the primary fails, failing-over to the standby might increase load and reduce performance,
- when a standby fails, the reads sent to the other standby (or the primary) can cause similar as above (or the reads on a standby might start failing).

So I would not advertise standby as 100% reliable and available "read-only" replicas.  This is another service.  But standbys can be used for low-availability or occasionnal read-only queries.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+    server
+
+## Failure handling
+
+### Minor failures
+
+Minor failures, such as service process crashes or temporary loss of
+network access, are handled automatically by Aiven in all plans without
+any major changes to the service deployment. The service automatically
+restores normal operation once the crashed process is restarted or when network access is
```
jfg-aiven, line 36: I would avoid the usage of the word restore as it could be confused with restoring a backup.  In this context, I would use the word resume.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+-   Typically reduces the data loss window in disaster scenarios
+-   Provides a quicker database time to restore with a controlled
+    failover in case of failures, as the standby is already installed,
+    running, and synchronised with the data
+-   Can be used for read-only queries to reduce the load on the primary
+    server
+
+## Failure handling
+
+### Minor failures
```
jfg-aiven, line 31: This section describes resuming service after a minor failure by restarting a crashed process.  In the presence of a standby node, I would not do this bvecause crash recovery of MySQL can take time, and then MySQL starts with a cold cache.  IMHO and when there is a standby node, I would prefer failing-over.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+Minor failures, such as service process crashes or temporary loss of
+network access, are handled automatically by Aiven in all plans without
+any major changes to the service deployment. The service automatically
+restores normal operation once the crashed process is restarted or when network access is
+restored.
+
+### Severe failures
+
+Severe failures, such as losing a node entirely in case of hardware
+or severe software problems, require involved recovery measures. The
```
jfg-aiven, line 42: I would give an example of software problem.  One such example is the unrecoverable crash-looping condition described in [Bug#111063](https://bugs.mysql.com/bug.php?id=111063).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+Aiven monitoring infrastructure automatically detects a failing node
+both when the node starts reporting issues in the self-diagnostics or
+when stops communicating. In such cases, the monitoring infrastructure
+automatically schedules a new replacement node to be created.
```
jfg-aiven, line 46: Right, a new replacement is scheduled, but when there is a standby, failover to standby is done, so I would add a note on this..

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Minor failures
+
+Minor failures, such as service process crashes or temporary loss of
+network access, are handled automatically by Aiven in all plans without
+any major changes to the service deployment. The service automatically
+restores normal operation once the crashed process is restarted or when network access is
+restored.
+
+### Severe failures
```
jfg-aiven, line 39: I would add RTO and RPO in this section (and probably in the minor section as well).  With standby, RTO is quick and RPO is close to zero loss (only loosing the unreplicated trx, and this could be a lossles if we implement semi-sync).  Wo standby, the RPO is lossing as much as 5 minutes of data (to be confirmed, we upload binlogs every 5 minutes) and the RTO is variable as it involves restoring a backup and applying binlogs which will depend on the db size and the write load (can be as much as 6 hours and even more in some cases).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+when stops communicating. In such cases, the monitoring infrastructure
+automatically schedules a new replacement node to be created.
+
+:::note
+In the event of database failover, the **Service URI** of your service
+remains the same; only the IP address will change to point to the new
+primary node.
+:::
+
+## Highly available Business and Premium service plans
```
jfg-aiven, line 54: It looks like above was for non HA plans, it was unclear to me, maybe a note should mention this.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+
+When the failed node is a MySQL standby node, the primary node
```
jfg-aiven, line 56: It is probably the place to mention that when the only standby is restored, a failure of the primary will be long to recover (restore a backup and apply binlogs, which can taker a long time).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+applications. Once the new replacement standby node is ready and
+synchronised with the primary node, it starts replicating the primary
+node in real time as the situation reverts back to normal.
+
+When the failed node is a MySQL primary node, the combined
+information from the Aiven monitoring infrastructure and the standby
+node is used to make a failover decision. The standby node is then
+promoted as the new primary and immediately starts serving clients. A
+new replacement node is automatically scheduled and becomes the new
+standby node. There is no data loss in this scenario.
```
jfg-aiven, line 67: I am not sure there is no data loss in this scenario as we do not implement semi-sync.  There is close to zero data loss, but some committed trx might not be replicated yet and are then loss.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+When the failed node is a MySQL standby node, the primary node
+keeps running normally and provides a normal service level to the client
+applications. Once the new replacement standby node is ready and
+synchronised with the primary node, it starts replicating the primary
+node in real time as the situation reverts back to normal.
+
+When the failed node is a MySQL primary node, the combined
+information from the Aiven monitoring infrastructure and the standby
+node is used to make a failover decision. The standby node is then
+promoted as the new primary and immediately starts serving clients. A
```
jfg-aiven, line 65: I would nuance "immediately" here.  If there is replication lag, this will not apply (or we might promote the node immediately, but then loose unapplied trx).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
+---
```
divdby0, line 1: My comments for the PostgreSQL HA page applies here as well.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+When the failed node is a MySQL primary node, the combined
+information from the Aiven monitoring infrastructure and the standby
+node is used to make a failover decision. The standby node is then
+promoted as the new primary and immediately starts serving clients. A
+new replacement node is automatically scheduled and becomes the new
+standby node. There is no data loss in this scenario.
+
+If all the primary and standby nodes fail at the same time or if all the primary
+nodes fail while standby nodes are being recovered, new
+nodes are automatically scheduled for creation to become the new primary
```
jfg-aiven, line 71: I agre with this in a single standby node plan.  But for a 2 standby node plan, I would not do a 2nd failover to the last standby before a new standby is provisioned.  This is to avoid a cascading failure condition that would lead to data lost.  And this is the subject of my March 2024 SRECon talk: [Autopsy of a Cascading Outage from a MySQL Crashing Bug](https://www.usenix.org/conference/srecon24americas/presentation/gagne).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+:::
+
+## Single-node Hobbyist and Startup service plans
+
+Hobbyist and Startup plans provide a single node; when it's lost, Aiven
+immediately starts the automatic process of creating a new replacement
+node. The new node starts up, restores its state from the latest
+available backup, and resumes serving customers.
+
+Since there is just a single node providing the service, the service is
```
divdby0, line 108: I think this paragraphs need a rewrite with engineering input. MySQL doesn't use WAL files (even though sometimes WAL is misused interchangeably).
cc: @alexole

divdby0 (changes requested):

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+promoted as the new primary and immediately starts serving clients. A
+new replacement node is automatically scheduled and becomes the new
+standby node. There is no data loss in this scenario.
+
+If all the primary and standby nodes fail at the same time or if all the primary
+nodes fail while standby nodes are being recovered, new
+nodes are automatically scheduled for creation to become the new primary
+and standby. The primary node is restored from the latest available
+backup, which can involve some degree of data loss. Any write
+operations made since the backup of the latest write-ahead logging (WAL) file are lost.
```
jfg-aiven, line 74: WAL are a Postgres notion.  The right notion in MySQL is the Binary Logs, which is different than the InnoDB Redo Logs (the Redo Logs are the WAL of MySQL, and these are different to the binlogs).

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+Typically, this time window is limited to either five minutes of time or one WAL file.
+
+If all the primary nodes fail while a standby is being recovered
```
jfg-aiven, line 77: Incomplete sentence.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+For Aiven for MySQL backups and restoration, Aiven uses
+[MyHoard](https://aiven.io/blog/introducing-myhoard-your-single-solution-to-mysql-backups-and-restoration).
+:::
+
+## Single-node Hobbyist and Startup service plans
+
+Hobbyist and Startup plans provide a single node; when it's lost, Aiven
+immediately starts the automatic process of creating a new replacement
+node. The new node starts up, restores its state from the latest
+available backup, and resumes serving customers.
```
jfg-aiven, line 106: Not only from the backup, but also applying the saved binlogs.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
...
+
+Since there is just a single node providing the service, the service is
+unavailable for the duration of the restoration. In addition, any write
+operations made since the backup of the latest WAL file are lost. Typically, this time
```
jfg-aiven, line 110: Not WAL, binlogs.

jfg-aiven (dismissed):
I added my comments.  Some of them should be checked by someone more familiar with our offer as my knowledge is limited there (I know MySQL very well, but I do not know Aiven's MySQL super well).

docs/products/postgresql/concepts/high-availability.md
jfg-aiven: I agreem I also made a similar comment.

docs/products/postgresql/concepts/high-availability.md
jfg-aiven: Right, no user involvment, but more time consuming as changing the DNS might take time (would be interesting to talk about this), and the presence of replication lag might also delay things and make things more complicated to recover.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,112 @@
+---
+title: High availability of Aiven for MySQL®
```
jfg-aiven, line 2: I rethought about this after doing my previous review.  In this doc, there are many references to backups and restores.  Maybe all the details about these should be put in a Backups and Restores page and a link to this new page should be put in here.  This might simplify the structure of this doc.

docs/products/mysql/concepts/high-availability-mysql.md
wojcik-dorota: I believe to this point it was neither HA nor non-HA, just generally applicable info.

docs/products/mysql/concepts/high-availability-mysql.md
wojcik-dorota: This is a good idea indeed. We might follow up on it in the future. For now, the bandwidth allows one new document only :-) Feel free to created a [DOC Jira](https://aiven.atlassian.net/jira/software/c/projects/DOC/boards/155) ticket for this future improvement. Thanks!

docs/products/mysql/concepts/high-availability-mysql.md
wojcik-dorota: @alexole , could you please comment?

docs/products/mysql/concepts/high-availability-mysql.md
jfg-aiven: DOC Jira done: https://aiven.atlassian.net/browse/DOC-858

docs/products/mysql/concepts/high-availability-mysql.md
alexole: For MySQL we do the fail-over in case of a crash loop, single crash does not lead to failover.

docs/products/mysql/concepts/high-availability-mysql.md
alexole: To me an extra section for RTO/RPO feels a bit of a duplication, as those are explained below.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,131 @@
...
+primary node.
+:::
+
+### RTO and RPO
+
+- With standby, recovery time objective (RTO) is quick, and recovery point objective (RPO)
+  is close to zero loss.
+- Without standby, RPO loses 5 minutes of data, and RTO is variable as it involves
+  restoring a backup and applying binlogs, which depends on the database size and the
+  write load (up to 6 hours or more in edge cases).
```
alexole, line 60: Not sure why we put *6* hours here. I'd say something like "up to several hours".

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,132 @@
+---
+title: High availability of Aiven for MySQL®
+sidebar_label: High availability
+---
+
+Aiven for MySQL® is available on a variety of plans, offering different levels of high availability. The selected plan defines the features available.
+
+| Plan         | High availability features                                      | Backup history | Uptime  |
```
wojcik-dorota, line 8: @alexole @divdby0 @jfg-aiven Could you please have a look at the `Uptime` quotas in the table. They are totally random, just added those we can get the correct values in the review. Do we have such data or shall I remove `Uptime` from the table?

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,132 @@
...
+of the standby nodes is the furthest along
+in replication (has the least potential for data loss) and does a
+controlled failover to that node.
+
+:::note
+For Aiven for MySQL backups and restoration, Aiven uses
+[MyHoard](https://aiven.io/blog/introducing-myhoard-your-single-solution-to-mysql-backups-and-restoration).
+:::
+
+## Single-node Hobbyist and Startup service plans
```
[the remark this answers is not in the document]
wojcik-dorota: @jfg-aiven And what specifically do you recommend to say?

docs/products/mysql/concepts/high-availability-mysql.md
jfg-aiven: I will let product answer this.

docs/products/mysql/concepts/high-availability-mysql.md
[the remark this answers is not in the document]
jfg-aiven: I do not have a suggestion.  I do not know what makes Startup better than Hobbyist, but with the uptime that were mentioned, I thought it was a natural question that deserved an answer.  Maybe @divdby0 can follow-up on this.

divdby0 (approved):

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,106 @@
...
+| ------------ | ------------------------------------------------------------------------------- | -------------- |
+| **Hobbyist** | Single node with limited availability                                           | 2 days         |
+| **Startup**  | Single node with limited availability                                           | 2 days         |
+| **Business** | Two nodes (primary + standby: with higher availability)                         | 14 days        |
+| **Premium**  | Three nodes (primary + 2 x standby: with top high availability characteristics) | 30 days        |
+
+## Primary and standby nodes
+
+Aiven's Business and Premium plans offer a primary node and standby nodes.
+A standby service is useful for multiple reasons:
```
staceysalamon-aiven, line 18:
```suggestion
A standby node is useful for multiple reasons:
```

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,106 @@
...
+any major changes to the service deployment. The service automatically
+restores normal operation once the crashed process is automatically restarted or when
+network access is restored.
+
+### Severe failures
+
+Severe failures, such as losing a node entirely in case of hardware
+or severe software problems, require radical recovery measures. The
+Aiven monitoring infrastructure automatically detects a failing node
+both when the node starts reporting issues in the self-diagnostics or
```
staceysalamon-aiven, line 44:
```suggestion
both when the node starts reporting issues in the self-diagnostics and
```

staceysalamon-aiven: Use "and" with "both", or remove "both" and keep "or".

docs/products/postgresql/concepts/high-availability.md
```diff
@@ -1,116 +1,121 @@
 ---
-title: High availability
+title: High availability of Aiven for PostgreSQL®
```
staceysalamon-aiven, line 2: As most (nearly all?) of the info on this page is duplicated in the general HA article, I wonder if it'd be better to just remove this and redirect there? The primary/standby node definitions in the terminology could also be added to the HA article instead of living in a separate page?

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,106 @@
+---
+title: High availability of Aiven for MySQL®
+sidebar_label: High availability
+---
+
+Aiven for MySQL® is available on a variety of plans, offering different levels of high availability. The selected plan defines the features available.
```
staceysalamon-aiven, line 6: I may be wrong, but I thought HA by definition didn't include single node services?

staceysalamon-aiven (commented):
Good info! I left some suggestions and questions. Also wondering if 2 articles are needed for this topic?

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,106 @@
...
+
+| Plan         | High availability features                                                      | Backup history |
+| ------------ | ------------------------------------------------------------------------------- | -------------- |
+| **Hobbyist** | Single node with limited availability                                           | 2 days         |
+| **Startup**  | Single node with limited availability                                           | 2 days         |
+| **Business** | Two nodes (primary + standby: with higher availability)                         | 14 days        |
```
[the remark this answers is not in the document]
wojcik-dorota: The working for describing types of high availability was specifically requested from the PM here so keeping that as is.

docs/products/mysql/concepts/high-availability-mysql.md
```diff
@@ -0,0 +1,106 @@
...
+| **Premium**  | Three nodes (primary + 2 x standby: with top high availability characteristics) | 30 days        |
```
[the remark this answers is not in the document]
wojcik-dorota: The wording for describing types of high availability was specifically requested from the PM here so keeping that as is.

docs/products/mysql/concepts/high-availability-mysql.md
wojcik-dorota: Here I guess we actually mean "service" meaning our standby node offering we have for out customers.

staceysalamon-aiven: Yeah, I understood it but thought it could potentially be confusing with how we use the term "service" within the platform. It's not a separate, distinct service, is it? Not sure if it would be misleading to a user or not.

docs/products/postgresql/concepts/high-availability.md
wojcik-dorota: Although indeed a huge part of this content can be shared by PG and MySQL, there are some technical details that actually differ these two articles, which is why reusing this content or cross-linking is not possible. Also, people are specifically requesting this type of info both for PG and MySQL so creating two separate articles seems the only way to address that temporarily before we have some better info architectural solutions for similar cases.

docs/products/mysql/concepts/high-availability-mysql.md
wojcik-dorota: ok, gotcha - since I know the context, it's not misleading for me but if it is to you I guess it might be also misleading for the users. changing to "note"...

staceysalamon-aiven (approved):
LGTM

ArthurFlag (changes requested):
Minor harmonization comment:  could you rename the file from `high-availability-mysql.md` to `high-availability.md`, that way we follow the existing pattern set by PG, and all the get-started.md, etc: no product name in the files, only for folders.

PR #133, opened 2024-03-25
update: group topics in Networking section

ArthurFlag:
## Description

Reduce number of topics by merging content where possible. Group remaining topic in the TOC.


staceysalamon-aiven (approved):
LGTM! The only thoughts I had were that maybe the VPC stuff could be grouped by cloud provider? (I have no idea if that makes sense for future updates, though.) And perhaps in another iteration the cloud regions and availability zones docs could be combined?

ArthurFlag:
Absolutely, that's my next step. I'll talk to SA in the next 2 weeks to determine the next level of grouping we need.

PR #141, opened 2024-03-29
update(platform): service integrations and restrict access with ip filter

wojcik-dorota:
https://aiven.atlassian.net/browse/AD-2258

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/platform/concepts/service-integration.md
+++ b/docs/platform/concepts/service-integration.md
@@ -2,15 +2,20 @@
 title: Service integrations
 ---
 
-Service integrations provide additional functionality and features by
-connecting different Aiven services together. This includes metrics
-integration which enables Aiven users to send advanced telemetry data to
-an [Aiven for PostgreSQL®](https://aiven.io/postgresql) database for
-metrics and visualize it in [Aiven for
-Grafana®](https://aiven.io/grafana). In addition to the metrics
-integration, log integration is supported that allows sending logs from
-any Aiven service to [Aiven for
-OpenSearch®](https://aiven.io/opensearch).
+Service integrations provide additional functionality and features by connecting different Aiven services together.
+
+## About service integrations
+
+Service integrations include metrics integration which enables Aiven users to send
+advanced telemetry data to an [Aiven for PostgreSQL®](https://aiven.io/postgresql)
+database for metrics and visualize it in [Aiven for Grafana®](https://aiven.io/grafana).
+In addition to the metrics integration, log integration is supported that allows sending
+logs from any Aiven service to [Aiven for OpenSearch®](https://aiven.io/opensearch).
+
+:::Tip
+For help with setting up a service integration or to request an integration type or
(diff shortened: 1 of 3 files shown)


docs/platform/howto/restrict-access.md
```diff
@@ -1,18 +1,22 @@
 ---
 title: Restrict network access to services
+sidebar_label: Restrict access
 ---
 
 import ActionsIcon from "@site/static/images/icons/more.svg";
 
 By default, Aiven services are publicly accessible, but you can restrict access to your service to a single IP, an address block, or any combination of both.
```
mlatief, line 8: I am thinking we can add here that if access from another service is needed, a service integration should be used, unless public access is allowed which is the default.

docs/platform/concepts/service-integration.md
```diff
@@ -2,15 +2,20 @@
...
-OpenSearch®](https://aiven.io/opensearch).
+Service integrations provide additional functionality and features by connecting different Aiven services together.
+
+## About service integrations
+
+Service integrations include metrics integration which enables Aiven users to send
+advanced telemetry data to an [Aiven for PostgreSQL®](https://aiven.io/postgresql)
+database for metrics and visualize it in [Aiven for Grafana®](https://aiven.io/grafana).
+In addition to the metrics integration, log integration is supported that allows sending
+logs from any Aiven service to [Aiven for OpenSearch®](https://aiven.io/opensearch).
```
mlatief, line 13: Would it make sense to also mention as an example the use of PG foreign data wrapper? Just to familiarize the audience with the IP Allow-list integrations.

wojcik-dorota: Sounds like a good idea indeed, but in this PR we're addressing specific issues as described in the ticket so maybe we should have a separate ticket for this one? Or do you prefer to add here?
In any case, please provide me with the details on the use of PG foreign data wrapper you'd like to add.

mlatief: I am afraid I don't have the best understanding of such details, but the more important point was to mention that a Service Integration should be used when connecting a service to another that's not publicly accessible :+1:

mlatief (approved):

docs/platform/howto/restrict-access.md
```diff
@@ -1,18 +1,27 @@
...
 
 :::important
-IP filters apply to publicly accessible endpoints only.
+IP filters apply to inbound traffic coming from specified IP addresses / network ranges to
+Aiven service ports.
 :::
 
+:::Tip
```
ArthurFlag, line 15: Let's stay as close to the [docusaurus docs](https://docusaurus.io/docs/markdown-features/admonitions) as possible, including casing, please 🙂 
```suggestion
:::tip
```

(applies to the whole PR, no just this file)

ArthurFlag (approved):

PR #151, opened 2024-04-04
update(postgresql): min & max client connection number per pool & service

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-950

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/postgresql/concepts/pg-connection-pooling.md
+++ b/docs/products/postgresql/concepts/pg-connection-pooling.md
@@ -18,22 +18,59 @@ usage on the server and better total performance.
 
 ## Maximum number of client connections
 
-Each pool can handle at least 5000 connections. This number increases for a pool by 500
-for each GB of RAM as per your service plan. To check the maximum number of client
-connections for your service, take into account the number of the connection pools and RAM
-GBs your service uses. Follow the sample calculation for a service using Business-4 plan
-with 4 GB of RAM and two connection pools:
+How many client connections your service can handle depends on the RAM size that your
+service plan supports.
 
-- Pool 1: 5000 + (500 x 4) = 7000
-- Pool 2: 5000 + (500 x 4) = 7000
-- Pool 1 + Pool 2 = 14000
+- Each gigabyte of RAM allows 500 connections.
+- Minimum number of client connections per service is 5000.
+- Maximum number of client connections per service is 50000.
 
-This Business-4 service can handle 14000 client connections.
+### Calculate `max_client_connections`
 
-:::note
-50000 is the maximum number of client connections per service and applies every time your
-calculated connection number exceeds 50000.
-:::
+Use the following formula to calculate how many client connections your service can handle:
+
+$$min(max(n * 500, 5000), 50000)$$
+
+Where:
+
+- `n` is the number of RAM GB that a service plan supports.
+- $$n * 500$$ is `intermedia_max_connections`.
+- $$5000 \leq intermedia\_max\_connections \leq 50000$$
+
+  - If `intermedia_max_connections` is less than 5000, lower bound 5000 applies.
+  - If `intermedia_max_connections` is greater than 50000, upper bound 50000 applies.
+
+### Examples
+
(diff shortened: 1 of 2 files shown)


ArthurFlag (approved):

docs/products/postgresql/concepts/pg-connection-pooling.md
```diff
@@ -2,28 +2,39 @@
...
-minimizing the consumption of server resources.
+Connection pooling in Aiven for PostgreSQL® services allows you to maintain very large numbers of connections to a database while minimizing the consumption of server resources.
+
+## About connection pooling
 
 Aiven for PostgreSQL connection pooling uses
 [PgBouncer](https://www.pgbouncer.org/) to manage the database
 connection.
 
-:::note
```
divdby0, line 13: I believe we should keep this original key information, perhaps again in a Node card without the explanation text (as it is explained below). Otherwise two key info is split by the example. 

- 500 for each GB of RAM in the service plan
-  A total of 50000 client connections

docs/products/postgresql/concepts/pg-connection-pooling.md
```diff
@@ -2,28 +2,39 @@
...
 PgBouncer automatically inserts the client queries and only uses a
 limited number of actual backend connections, leading to lower resource
 usage on the server and better total performance.
 
+## Maximum number of client connections
+
+Each pool can handle at least 5000 connections. This number increases for a pool by 500
+for each GB of RAM as per your service plan. To check the maximum number of client
+connections for your service, take into account the number of the connection pools and RAM
+GBs your service uses. Follow the sample calculation for a service using Business-4 plan
```
divdby0, line 24: I suggest splitting the paragraph from the Follow the sample part and make the latter more obvious as an example.

docs/products/postgresql/concepts/pg-connection-pooling.md
wojcik-dorota: Instead of duplicating this info in two sections, I reorganized the other section so that the key info is no longer split by the example. Please have a look. Wdyt?

divdby0: Thanks. I think this version is much better as it retains the key information together, more visible but it also explains supported by an example. 👌🏻

The only remaining point for me is that 50.000 is the limit per service and there is no way of exceeding it. You might consider dropping the part "and applies every time your calculated connection number exceeds 50000."

divdby0 (approved):

docs/products/postgresql/concepts/pg-connection-pooling.md
```diff
@@ -2,28 +2,38 @@
...
+## Maximum number of client connections
+
+To check the maximum number of client connections for your service, take into account the
+number of the connection pools and RAM GBs your service uses, and apply the following rules:
+
+- Each pool can handle at least 5000 connections.
+- This number (5000) increases for a pool by 500 for each GB of RAM as per your service plan.
+- 50000 is the maximum number of client connections per service.
+
+For example, if a service using Business-4 plan with 4 GB of RAM has two connection pools,
```
divdby0, line 28: @0xlianhu I would appreciate if you could validate this example. Thanks.

0xlianhu: We cannot release this document until we fix the bug I mentioned in the slack channel.

0xlianhu: According to others' discussion and existing test cases, I think the document should be updated to:
minimum 5000, max 50000. 

Hence,  for a 4G machine, it should be 5000 instead of 7000. 

I.e., 
        (1024 * 2, 5000),
        (1024 * 4, 5000),
        (1024 * 8, 5000),
        (1024 * 16, 8000),
        (1024 * 32, 16000),
        (1024 * 64, 32000),
        (1024 * 128, 50000),
        (1024 * 256, 50000),

0xlianhu (approved):

PR #172, opened 2024-04-29
add(platform): byoc avn client commands

wojcik-dorota:
https://aiven.atlassian.net/browse/MILK-115

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/tools/cli/byoc.md
+++ b/docs/tools/cli/byoc.md
@@ -1,0 +1,272 @@
+---
+title: avn byoc
+---
+
+Set up and manage your [custom clouds](/docs/platform/concepts/byoc) using the Aiven client and `avn byoc` commands.
+
+## Manage a custom cloud
+
+### `avn byoc create`
+
+[Creates a custom cloud](/docs/platform/howto/byoc/create-custom-cloud) in an organization.
+
+| Parameter           | Required | Information                                                                                                           |
+| ------------------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
+| `--organization-id` | Yes      | Identifier of an organization where to create a custom clouds                                                         |
+| `--deployment-model`| Yes      | Private or public network architecture model defining how resources are arranged and connected to your cloud provider |
+| `--cloud-provider`  | Yes      | Cloud provider to be used for running a custom cloud                                                                  |
+| `--cloud-region`    | Yes      | Cloud region where to create a custom cloud                                                                           |
+| `--reserved-cidr`   | Yes      | IP address range of the VPC to be created in your cloud account for Aiven services hosted on a custom cloud           |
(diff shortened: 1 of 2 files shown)


docs/tools/cli/byoc.md
```diff
@@ -0,0 +1,127 @@
...
+
+### `avn byoc cloud permissions add`
+
+Grants permissions to a custom cloud.
+
+| Parameter           | Required | Information                                                                                                           |
+| ------------------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
+| `--organization-id` | Yes      | Identifier of an organization where a custom cloud is located                                                         |
+| `--byoc-id`         | Yes      | Identifier of a custom cloud for which to grant permissions                                                           |
+| `--account`         | No       | Identifier of your account                                                                                            |
```
heikju, line 73: This is used for giving the identifier of an account (organizational unit), so that the custom cloud is available in all projects belonging to that account. The description does not make that quite clear I think. In other words, this also grants permission to use the cloud like the option below, but with a different scope.

And the same goes for any instances of the `--account` option in other commands' descriptions.

docs/tools/cli/byoc.md
```diff
@@ -0,0 +1,127 @@
...
+Downloads a custom cloud Terraform template.
+
+| Parameter           | Required | Information                                                             |
+| ------------------- | -------- | ----------------------------------------------------------------------- |
+| `--organization-id` | Yes      | Identifier of an organization where a custom cloud is located           |
+| `--byoc-id`         | Yes      | Identifier of a custom cloud for which to download a Terraform template |
+
+### `avn byoc templete terraform get-vars`
+
+Downloads a custom cloud variables file.
```
heikju, line 122: We could call it a "Terraform variables file" just to be more precise and match TF's own documentation.

docs/tools/cli/byoc.md
```diff
@@ -0,0 +1,127 @@
...
+| `--organization-id` | Yes      | Identifier of an organization where to delete a custom cloud            |
+| `--byoc-id`         | Yes      | Identifier of a custom cloud to be deleted                              |
+
+### `avn byoc list`
+
+Returns a list of all the custom clouds in an organization.
+
+| Parameter          | Required | Information                                                              |
+| ------------------ | -------- | ------------------------------------------------------------------------ |
+| `--organization-id`| Yes      | Identifier of an organization where to check for available custom clouds |
```
heikju, line 37: I think "check for available custom clouds" is a bit unclear and IMO not needed. "Identifier of the organization" would be enough of a description in my opionion.

heikju (dismissed):
Some comments, but nothing drastic. This kind of a brief reference seems like a decent starting point. Adding some realistic examples could be a good idea later, but not needed at this point.

heikju (approved):

ArthurFlag (approved):

PR #186, opened 2024-05-07
Remove region specificity from free tier docs

webchick:
It was raised at https://aiven.io/community/forum/t/free-plan-postgresql-getting-moved-to-digital-ocean-docs-say-aws/751/3?u=nikkee that our list of free tier servers is no longer accurate. This strikes me as the type of information that's at risk of falling out of date periodically, so one suggestion is to remove the specificity.

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/platform/concepts/free-plan.md
+++ b/docs/platform/concepts/free-plan.md
@@ -34,11 +34,9 @@ Free plans include:
 -   Backups
 -   Integrations between different Aiven services including free, paid,
     and trial services
--   DigitalOcean hosting in a limited number of regions:
-    -   EMEA: do-ams (Amsterdam), do-ldn (London), do-fra (Frankfurt)
-    -   Americas: do-nyc (New York), do-sfo (San Francisco), do-tor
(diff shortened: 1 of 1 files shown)


staceysalamon-aiven:
> It was raised at https://aiven.io/community/forum/t/free-plan-postgresql-getting-moved-to-digital-ocean-docs-say-aws/751/3?u=nikkee that our list of free tier servers is no longer accurate. This strikes me as the type of information that's at risk of falling out of date periodically, so one suggestion is to remove the specificity.


This was my suggestion when the original documentation was created with the AWS regions and there was a case made for including them. I will try to find the Slack thread for this.

webchick:
Actually the original poster made a case for not obfuscating this information as well: https://aiven.io/community/forum/t/free-plan-postgresql-getting-moved-to-digital-ocean-docs-say-aws/751/5?u=aiven_angie

So I think I will "won't fix" this one and open a separate PR that solves the actual problem.

PR #193, opened 2024-05-13
fix: icons for users, app users, and groups

staceysalamon-aiven:
## Describe your changes

* Fixed the users icon
* Fixed the copy for the application users icon
* Added Groups

## Checklist

- [X] I checked the output.

### DIFF
--- a/src/components/ConsoleIcons/index.tsx
+++ b/src/components/ConsoleIcons/index.tsx
@@ -47,6 +47,13 @@ export default function ConsoleLabel({name}): ReactElement {
           <b>Metrics</b>
         </>
       );
+    case 'serviceusers':
+      return (
+        <>
+          <ConsoleIconWrapper icon={ConsoleIcons.people} />{' '}
+          <b>Users</b>
+        </>
+      );
     case 'services':
       return (
         <>
@@ -135,6 +142,13 @@ export default function ConsoleLabel({name}): ReactElement {
           <ConsoleIconWrapper icon={ConsoleIcons.user} /> <b>Users</b>
         </>
       );
+    case 'viewuserprofile':
+      return (
+        <>
+          <ConsoleIconWrapper icon={ConsoleIcons.user} />{' '}
+          <b>View profile</b>
+        </>
+      );
     case 'groups':
       return (
         <>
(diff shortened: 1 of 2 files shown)


src/components/ConsoleIcons/index.tsx
```diff
@@ -132,7 +132,13 @@ export default function ConsoleLabel({name}): ReactElement {
     case 'users':
       return (
         <>
-          <ConsoleIconWrapper icon={ConsoleIcons.people} /> <b>Users</b>
+          <ConsoleIconWrapper icon={ConsoleIcons.user} /> <b>Users</b>
```
ArthurFlag, line 135: That's not what I experience in the sidebar of any service?

Screenshot 2024-05-13 at 12 42 33

Current state seems to be correct for that area. Where did you see the `users` label connected to a single user icon? We might need to add a case.

ArthurFlag (changes requested):

src/components/ConsoleIcons/index.tsx
staceysalamon-aiven: Ah, okay, so this is service users? Org users are managed in Admin. 
image

staceysalamon-aiven: Speaking of which, I've just noticed as well that the view profile icon is different for org users versus applications users.

staceysalamon-aiven: Added cases for these 👍

src/components/ConsoleIcons/index.tsx
```diff
@@ -198,10 +218,11 @@ export default function ConsoleLabel({name}): ReactElement {
     case 'applicationusers':
       return (
         <>
-          <ConsoleIconWrapper icon={ConsoleIcons.appUsers} /> <b>Users</b>
+          <ConsoleIconWrapper icon={ConsoleIcons.appUsers} />{' '}
+          <b>Application users</b>
         </>
       );
-    case 'viewprofile':
+    case 'viewappuserprofile':
```
ArthurFlag, line 225: Have you checked if it's used already? If so, the files using this case must be updated :)

staceysalamon-aiven: Not from what I can see, unless I'm missing something?

src/components/ConsoleIcons/index.tsx
staceysalamon-aiven: Missed the service users page, but asking the team about this since it doesn't seem totally accurate?

PR #222, opened 2024-05-28
Added instructions for completing VPC peering in UpCloud via the web console

kevinbowman-aiven:
## Describe your changes

UpCloud sent us instructions for VPC peering via the web console, previously we only had instructions for using their API.

## Checklist

(this is an update rather than a new page, so most of these were already in place)

- [x] I checked the output.


wojcik-dorota (approved):
Approving with some minor editorial updates. Thank you for your contribution ❤️

PR #238, opened 2024-06-04
update(clickhouse): more clarity for users and roles

wojcik-dorota:
https://aiven.atlassian.net/browse/DDB-1045

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/clickhouse/howto/manage-users-roles.md
+++ b/docs/products/clickhouse/howto/manage-users-roles.md
@@ -1,21 +1,20 @@
 ---
-title: Manage ClickHouse® service users and roles
+title: Manage Aiven for ClickHouse® users and roles
+sidebar_label: Manage users and roles
 ---
 
-This article shows you how to create new user accounts for your service
-and how to manage user roles and permissions for efficiently access
-control.
+Create Aiven for ClickHouse® users and roles and grant them specific permissions to efficiently control or restrict access to your service.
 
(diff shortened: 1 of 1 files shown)


docs/products/clickhouse/howto/manage-users-roles.md
```diff
@@ -24,86 +23,93 @@ To create a user account for your service,
...
 
 ### Grant permissions
 
 You can grant permissions both to specific roles and to individual
 users. The grants can be also granular, targeting specific databases,
 tables, columns, or rows.
 
-For example, the following request grants the `auditor` role permissions
+:::important
+You cannot grant privileges to the main service user. Aiven may grant privileges to the
```
kmichel-aiven, line 76:
```suggestion
You cannot grant additional privileges to the main service user. Aiven may grant privileges to the
```

kmichel-aiven (commented):
The changes looks good to me.

Looking at the whole page, we should probably be consistent and use the same word instead of alternating between "privileges" and "permissions" (they are the same thing, using two words for the same concept might be confusing and suggest otherwise).

Looking a the upstream documentation, they are also jumping between the two words, but it looks like "privileges" is used more often.

What do you think ?

kmichel-aiven (approved):

harshini-rangaswamy (approved):
LGTM :rocket:

PR #334, opened 2024-07-17
update: use cards where possible

ArthurFlag:
Delete duplicated getting started and use cards on landing pages.


wojcik-dorota (approved):
LGTM

PR #339, opened 2024-07-19
update: remove overview section

ArthurFlag:
Harmonize with other services by having "Get started" as first page, and increase value of the overview page.


wojcik-dorota (approved):
LGTM

PR #423, opened 2024-08-28
update: security info for personal token auth policy

staceysalamon-aiven:
## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/platform/howto/set-authentication-policies.md
+++ b/docs/platform/howto/set-authentication-policies.md
@@ -47,29 +47,37 @@ strongly recommended to enable this if you only have one Aiven organization.
 
 ### Personal tokens
 
-Users can generate their own
-[personal tokens](/docs/platform/howto/create_authentication_token) for use with
-the Aiven API. When you turn off personal tokens, managed users can't create
-personal tokens. Non-managed users can still create personal tokens, but they can't use
-them to access the organization's resources.
-
-To regularly manage your resources programmatically with the API, CLI, Terraform,
-or other applications, it's best to create an [application user](/docs/platform/howto/manage-application-users) with its own tokens.
-
-Personal tokens are generated with the authentication method that the user logged in with.
-Tokens are linked to the authentication method they are created with. You can ensure that
-access to your organization using tokens conforms to the authentication policy by
-requiring users to be logged in with an allowed authentication method when they
-use a token. If your authentication policy changes, tokens that don't conform to the
+[Personal tokens](/docs/platform/howto/create_authentication_token) are tokens that
+users create themselves for use with the Aiven API. Personal tokens are generated with
+and linked to the authentication method that the user logged in with. You can ensure
+that access to your organization using tokens conforms to the authentication policy
+by requiring users to be logged in with an allowed authentication method when
+they use a token. If your authentication policy changes, tokens that don't conform to the
 new policy stop working.
 
 For example, if you have an authentication policy that allows users to log in with a
 password, a user can log in with their email and password, and create a personal token.
 This token is tied to the password authentication method they logged in with.
-If the authentication policy changes later to only allow logging on with an identity
+If the authentication policy changes later to only allow logging in with an identity
 provider, then the token generated when the user was logged in with their password will
-not work. After logging in with an allowed method on the new authentication policy
+not work. After logging in with the allowed method on the new authentication policy
 the user can create a token.
+
+The authentication policy setting for personal tokens only applies to managed users.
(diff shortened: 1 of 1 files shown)


docs/platform/howto/set-authentication-policies.md
```diff
@@ -47,30 +47,38 @@ strongly recommended to enable this if you only have one Aiven organization.
...
 This token is tied to the password authentication method they logged in with.
-If the authentication policy changes later to only allow logging on with an identity
+If the authentication policy changes later to only allow logging in with an identity
 provider, then the token generated when the user was logged in with their password will
-not work. After logging in with an allowed method on the new authentication policy
+not work. After logging in with the allowed method on the new authentication policy
 the user can create a token.
 
+The authentication policy setting for personal tokens only applies to managed users.
+If **Allow managed users to create personal tokens** is disabled, managed users
```
italomg, line 67: We should probably change the text in the console before updating the docs, I can create a ticket for that in case it doesn't exist.

docs/platform/howto/set-authentication-policies.md
```diff
@@ -47,30 +47,38 @@ strongly recommended to enable this if you only have one Aiven organization.
...
+cannot create personal tokens. Non-managed users can still create personal tokens,
+but they can't use them to access the organization's resources.
+
+Even with this setting disabled, managed users can still create personal tokens
```
italomg, line 71: when the setting is disabled managed users shouldn't be able to create personal tokens. However, if they have personal tokens before becoming managed users then they can still use those personal tokens for doing operations in other organizations besides the one that manages them(assuming those orgs allow personal tokens).

docs/platform/howto/set-authentication-policies.md
```diff
@@ -47,30 +47,38 @@ strongly recommended to enable this if you only have one Aiven organization.
...
+in any other organization they are a member of that has it enabled. They can
+use the personal tokens from other organizations to access your organization's
+resources. To restrict the use of personal tokens from other organizations, select
```
italomg, line 74: > They can use the personal tokens from other organizations to access your organization's resources.

this sounds quite bad :disappointed: . The personal tokens are not associated with any organization per se, so maybe better say "They can use their personal tokens to access your organization's resources." or something along those lines.

docs/platform/howto/set-authentication-policies.md
```diff
@@ -47,30 +47,38 @@ strongly recommended to enable this if you only have one Aiven organization.
...
 the user can create a token.
 
+The authentication policy setting for personal tokens only applies to managed users.
+If **Allow managed users to create personal tokens** is disabled, managed users
+cannot create personal tokens. Non-managed users can still create personal tokens,
+but they can't use them to access the organization's resources.
+
+Even with this setting disabled, managed users can still create personal tokens
+in any other organization they are a member of that has it enabled. They can
+use the personal tokens from other organizations to access your organization's
```
italomg, line 73: > They can use the personal tokens from other organizations to access your organization's resources.

this sounds quite bad :disappointed: . The personal tokens are not associated with any organization per se, so maybe better say "They can use their personal tokens to access your organization's resources." or something along those lines.

docs/platform/howto/set-authentication-policies.md
italomg, line 74: > To restrict the use of personal tokens from other organizations, select **Require users to be logged in with an allowed authentication method**.

Considering my previous comment we might want to changes this also, maybe "To enforce that personal tokens used conform with your organization authentication policy select
**Require users to be logged in with an allowed authentication method**."

but then again, we should probably reword the option in the console :thinking:

docs/platform/howto/set-authentication-policies.md
staceysalamon-aiven: I had thought we were going to keep Console copy the same, but agree that we should do that first and make it clearer! 💯

docs/platform/howto/set-authentication-policies.md
staceysalamon-aiven: > if they have personal tokens before becoming managed users
Is this referring to personal tokens that were created in any organization?

docs/platform/howto/set-authentication-policies.md
staceysalamon-aiven: It does, yeah, but I wanted to be clear that the _ability_ to create tokens is linked to the org auth policy. I'll figure out a way to reword this to make it clear that it's still a _personal_ token and not actually associated with the org, though.

PR #466, opened 2024-09-17
add: content for ISM policies after snapshot restore

harshini-rangaswamy:
## Checklist

- [ ] I checked the output.


docs/products/opensearch/howto/migrate-ism-policies.md
```diff
@@ -0,0 +1,104 @@
...
+- **Script can only be run once**
+  OpenSearch clears the cluster metadata with ISM policy assignments after the policies
+  are applied. This means the script can only be run once. To run the script again,
+  restore the snapshot.
+
+:::
+
+## Check for active ISM policies
+
+Before executing the reapply script, confirm that no indices are currently managed by
```
mteiste, line 44: The script does this so no separate step is needed.

docs/products/opensearch/howto/migrate-ism-policies.md
```diff
@@ -0,0 +1,104 @@
...
+ISM policies. Run the following command to check:
+
+```bash
+curl -X GET --insecure -s "${AIVEN_SERVICE_URI}/_plugins/_ism/explain?pretty"
+```
+
+If the result shows `"total_managed_indices": 0`, no policies are applied, and you can
+proceed. Otherwise, remove the existing policies if necessary.
```
mteiste, line 52: This is not the recovery, e.g. removing the policies will help if the script has already been run targeting the service as the metadata it uses is no longer available.

harshini-rangaswamy: Thanks. Updated the text to clarify that removing policies is only needed if the script has already been run and the metadata is unavailable. Also, moved this into the next section.

mteiste (approved):

docs/products/opensearch/howto/migrate-ism-policies.md
```diff
@@ -0,0 +1,91 @@
...
+   ```json
+   {
+       "host": "target-ip-or-fqdn",
+       "port": target-port-number,
+       "user": "the user",
+       "password": "the password"
+   }
+   ```
+
+   Use the default `aivenadmin` user for this task.
```
ArthurFlag, line 63: I'm not sure I understand this instruction, is it related to line 58, or to this task in general (line  41) ?

harshini-rangaswamy: Good catch. It is the 'user' in the config file. Updated step to clarify this.

ArthurFlag (approved):

PR #472, opened 2024-09-20
update(platform): ip filter

wojcik-dorota:
https://aiven.atlassian.net/browse/MILK-276

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/platform/howto/restrict-access.md
+++ b/docs/platform/howto/restrict-access.md
@@ -5,12 +5,20 @@ sidebar_label: Restrict access
 
 import ConsoleLabel from "@site/src/components/ConsoleIcons"
 
-By default, Aiven services are publicly accessible, but you can restrict access to your service to a single IP, an address block, or any combination of both.
+Restrict access to your Aiven-managed service to a single IP, an address block, or any combination of both.
 
-:::important
-IP filters apply to inbound traffic coming from specified IP addresses / network ranges to
-Aiven service ports.
+By default, a connection to an Aiven service can be established from any IP address. To
+restrict access, you can use the IP filtering capability, which allows you to filter
+traffic incoming to your services by specifying allowed IP addresses / network ranges.
+
+:::note
+If your service is within a VPC, the VPC configuration filters incoming traffic before the
+IP filter is applied.
 :::
+
+By default, the IP filter is set `0.0.0.0/0`, which allows all inbound connections. If you
+remove `0.0.0.0/0` without adding networks or addresses used by clients, no client can
+connect to your service.
 
 :::tip
 To access a non-publicly-accessible service from another service, use a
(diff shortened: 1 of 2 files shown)


docs/platform/howto/restrict-access.md
```diff
@@ -5,40 +5,42 @@ sidebar_label: Restrict access
...
+you can use the IP filtering capability, which allows you to filter traffic incoming to
+your services by specifying allowed IP addresses / network ranges.
+
+note:::
+If your service is within a VPC, the VPC configuration filters incoming traffic before the
+IP filter is applied.
 :::
 
+By default, the IP filter is set `0.0.0.0/0`, which allows all inbound connections. If you
+remove `0.0.0.0/0`, your service loses all connectivity, including private network addresses.
```
aihessu, line 20: ... If you remove 0.0.0.0/0 without adding the networks or addresses used by the clients, no clients will be able to connect.

(The service does not loose "all connectivity" since many services can still do outbound connections, integrations etc; the filter only applies to client connections. The 0.0.0.0/0 also needs to be removed when adding the actual correct client networks in the ACL, otherwise all addresses are still allowed to connect.

docs/platform/howto/restrict-access.md
```diff
@@ -5,40 +5,42 @@ sidebar_label: Restrict access
...
+1. On the <ConsoleLabel name="overview"/> page of your service, select
+   <ConsoleLabel name="service settings"/>.
+1. On the <ConsoleLabel name="service settings"/> page, in the **Cloud and
    network** section, click <ConsoleLabel name="actions"/> > **Set public IP filters**.
-
-1. In the **Allowed inbound IP addresses** window, enter your address
-   or address block using the CIDR notation, and select the **+** icon
-   to add it to the list of the trusted IP addresses.
+1. In the **Allowed inbound IP addresses** window, click **Add IP address range**, and
+   enter your address or address block using the CIDR notation, for example `10.20.0.0/16`.
```
aihessu, line 36: If the customer adds a new IP range like this, and does not remove the default 0.0.0.0/0 (which allows everything), the service is still wide open to the world, and the change has no effect.

When adding the first subnet there, the default of 0.0.0.0/0 must be replaced with 10.20.0.0/16, then click "Save changes". If additional subnets beyond that need to be allowed, the "Add IP address range" button must be clicked.

aihessu (dismissed):

docs/platform/howto/restrict-access.md
```diff
@@ -5,40 +5,49 @@ sidebar_label: Restrict access
 
 import ConsoleLabel from "@site/src/components/ConsoleIcons"
 
-By default, Aiven services are publicly accessible, but you can restrict access to your service to a single IP, an address block, or any combination of both.
+Restrict access to your Aiven-managed service to a single IP, an address block, or any combination of both.
 
-:::important
-IP filters apply to inbound traffic coming from specified IP addresses / network ranges to
-Aiven service ports.
+By default, Aiven services are publicly accessible. To restrict access to Aiven services,
```
aihessu, line 10: Just realised that it's a bit too much to say that the services are publicly accessible by default. There is user/client authentication in place, so the service itself is not publicly accessible. But by default, connections to the service can be made from any address, after which the authentication takes place. But maybe this is obvious to the reader.

aihessu (approved):

docs/platform/howto/restrict-access.md
```diff
@@ -5,40 +5,49 @@ sidebar_label: Restrict access
...
 
-By default, Aiven services are publicly accessible, but you can restrict access to your service to a single IP, an address block, or any combination of both.
+Restrict access to your Aiven-managed service to a single IP, an address block, or any combination of both.
 
-:::important
-IP filters apply to inbound traffic coming from specified IP addresses / network ranges to
-Aiven service ports.
+By default, a connection to an Aiven service can be established from any IP address. To
+restrict access, you can use the IP filtering capability, which allows you to filter
+traffic incoming to your services by specifying allowed IP addresses / network ranges.
```
staceysalamon-aiven, line 12: Suggestion to remove the slash

```suggestion
traffic incoming to your services by specifying allowed IP addresses or network ranges.
```

docs/platform/howto/restrict-access.md
```diff
@@ -5,40 +5,49 @@ sidebar_label: Restrict access
...
+
+:::note
+If your service is within a VPC, the VPC configuration filters incoming traffic before the
+IP filter is applied.
 :::
 
+By default, the IP filter is set `0.0.0.0/0`, which allows all inbound connections. If you
```
staceysalamon-aiven, line 19:
```suggestion
By default, the IP filter is set to `0.0.0.0/0`, which allows all inbound connections. If you
```

staceysalamon-aiven (approved):
One small suggestion and a super minor typo fix. Otherwise, LGTM 👍

PR #479, opened 2024-09-25
add: anchors for advanced params

ArthurFlag:
Example: https://doc-1180-add-permalink-to-ad.aiven-docs.pages.dev/docs/products/dragonfly/reference/advanced-params


ArthurFlag:
> Could you please also apply the same on this PR https://github.com/aiven/aiven-docs-internal/pull/96/files ?

I'll be automatically apply when we release :)

harshini-rangaswamy:
> Example: https://1da4d579.aiven-docs.pages.dev/docs/products/mysql/reference/advanced-params#service_log
This is cool. Also thinking, instead of linking the `#` at the end, would it be possible to move the `#` before the parameter? Maybe we could wrap the entire parameter name in the `<a>` tag, making the name linkable/clickable rather than just the `#` symbol. This would provide a more intuitive experience. I'm not sure how feasible this is, though (just thinking out loud).

ArthurFlag:
I'm essentially copying Docusaurus' style by keeping the hashtag at the end but yeah I can tweak it to make the whole first line clickable.

harshini-rangaswamy (approved):
This looks good :rocket:. Thanks!

PR #520, opened 2024-10-22
fix: `avn service cli` also support Valkey

TibsAtWork:
## Describe your changes

`avn service cli` also supports connecting to an Aiven for Valkey service.

It connects to Valkey or Redis using whichever of `valkey-cli` or `redis-cli` is available (validated by trying it out!), but I think it's more useful to reference `valkey-cli` (only) in the text (it does say "such as"), especially as that will make it simpler to remove mention of Aiven for Caching when that is retired.

## Checklist

- [x] I checked the output.


staceysalamon-aiven (approved):
Thanks for doing this!

PR #522, opened 2024-10-22
fix: free plan page should name Valkey

TibsAtWork:
We no longer support a free plan for Aiven for Caching, we do for Aiven for Valkey, so replace the terms.

Note: I assume that the `maxmemory` limitation is retained as-is.

## Checklist

- [x] I checked the output.


harshini-rangaswamy (approved):
Thanks for doing this.

PR #557, opened 2024-11-11
add(clickhouse): total storage limitation

wojcik-dorota:
https://aiven.atlassian.net/browse/DDB-1361

## Checklist

- [x] I checked the output.

### DIFF
--- a/static/includes/clickhouse-storage-limitation.md
+++ b/static/includes/clickhouse-storage-limitation.md
@@ -1,0 +1,9 @@
+:::note[Total storage per plan]
+In service plans with a single shard, storage capacity doesn't depend on the number of
+nodes, and the data of this single shard is replicated on all the available nodes.
+
+For example, with a Startup-16 plan you get 1150 GB of total storage per VM, same as with
+a Business-16 plan. Since the Business-16 plan offers three VMs, your total storage is
+3450 GB, but effectively it's still 1150 GB because that’s the maximum a single node can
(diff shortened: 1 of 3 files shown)


static/includes/clickhouse-storage-limitation.md
```diff
@@ -0,0 +1,9 @@
+:::note[Total storage per plan]
+In service plans with a single shard, storage capacity doesn't depend on the number of
+nodes, and the data of this single shard is replicated on all the available nodes.
+
+For example, with a Startup-16 plan you get 1150 GB of total storage per VM, same as with
```
Khatskevich, line 5: That is a good explanation.
I would also add a small clarification for the sharded plan here.

wojcik-dorota: Thank you @Khatskevich ! What would you specifically include for the sharded plan?

Khatskevich: May be give another example: For example we have Premium-6x-16 which has 2 shards and 6 servers.
This means that there are 6 servers with 1150GB storage each and the data is replicated 3 times.

And the conclusion: Total storage is the amount of data which may be inserted to the service, which will be replicated `number of nodes` / `number of shards` times.

Khatskevich (approved):

harshini-rangaswamy (approved):
Look good!

PR #591, opened 2024-12-09
Dorota clickhouse add query cache

wojcik-dorota:
https://aiven.atlassian.net/browse/DDB-1502

## Checklist

- [ ] I checked the output.


docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,82 @@
...
+## Why use it
+
+Using query cache in your Aiven for ClickHouse services can help reduce latency and
+resource consumption.
+
+Key use case for the Aiven for ClickHouse query cache are the following:
+
+- Performance enhancement: Reducing latency and load for frequently executed queries
+- Analytical workloads: Using complex repetitive queries that involve high data aggregation
+- Data pipelines: [Using materialized views](/docs/products/clickhouse/concepts/federated-queries)
```
Khatskevich, line 31: I do not understand how materialized views are related to the Query Cache. Looks like it is actually another way to handle the problem

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,82 @@
...
+  to optimize recurring data transformation and aggregation
+
+## Enable query cache
+
+To enable the query cache for a query, use an SQL client (for example, the
+[ClickHouse client](/docs/products/clickhouse/howto/connect-with-clickhouse-cli)) and
+append `SETTINGS use_query_cache = 1` to your query, for example:
```
Khatskevich, line 38: Looks like it is the only way to use the query cache. In fact you need to change the setting and it is the just easiest way to do so.

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,82 @@
...
+- `query_cache_ttl`
+- `query_cache_share_between_users`
+
+## Limitation
+
+- Maximum query cache size: 64 MiB for each GiB of RAM (for example, 256 MiB for a 4-GiB
+  service or 1 GiB for a 16-GiB service)
+- Maximum number of query cache entries: 64 entries for each GiB of RAM (for example,
+  1024 entries for a 16-GiB service)
+- Query cache is not long-lived: cached results are invalidated when the underlying data
```
Khatskevich, line 74: No, if data changes, the cached result is not invalidated and may return incorrect result https://clickhouse.com/blog/introduction-to-the-clickhouse-query-cache-and-design

Khatskevich (changes requested):

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,76 @@
...
+resource consumption.
+
+Key use case for the Aiven for ClickHouse query cache are the following:
+
+- Performance enhancement: Reducing latency and load for frequently executed queries
+- Analytical workloads: Using complex repetitive queries that involve high data aggregation
+
+## Enable query cache
+
+To enable the query cache for a query, use an SQL client (for example, the
```
Khatskevich, line 32: It still sounds like using client and applying settings in this way is necessary.
I would white something like
"To enable the query cache, setting `use_query_cache` should be set to `1`. It can be done by adding `SETTINGS use_query_cache = 1` to the end of your query.

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,76 @@
...
+- `query_cache_squash_partial_results`
+- `query_cache_ttl`
+- `query_cache_share_between_users`
+
+## Limitation
+
+- Maximum query cache size: 64 MiB for each GiB of RAM (for example, 256 MiB for a 4-GiB
+  service or 1 GiB for a 16-GiB service)
+- Maximum number of query cache entries: 64 entries for each GiB of RAM (for example,
+  1024 entries for a 16-GiB service)
```
Khatskevich, line 69: The main limitation is that the query is not updated when the underlying data changes

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,79 @@
...
+---
+
+Aiven for ClickHouse® provides a query cache mechanism that helps improve query performance
+by caching query results.
+
+## How it works
+
+When the Aiven for
+[ClickHouse query cache](https://clickhouse.com/docs/en/operations/query-cache) is enabled,
+multiple identical `SELECT` queries run simultaneously are computed only once,
```
harshini-rangaswamy, line 13: Could we consider revising this for clarity? For example:
```suggestion
multiple identical `SELECT` queries run simultaneously and are computed only once,
```
```suggestion
multiple identical `SELECT` queries running simultaneously are computed only once
```

docs/products/clickhouse/howto/clickhouse-query-cache.md
```diff
@@ -0,0 +1,79 @@
...
+and further executions of the same query are served directly from the cache.
```
harshini-rangaswamy, line 14: How about this, break the sentence: 
```suggestion
. Subsequent executions of the same query are served directly from the cache.
```

..identical SELECT queries running simultaneously are computed only once. Subsequent executions of the same query are served directly from the cache.

harshini-rangaswamy (approved):
Looks good to me :rocket:. Left some minor suggestions :)

PR #593, opened 2024-12-11
add: IP address range feature for tokens and auth policy

staceysalamon-aiven:
## Checklist

- [X] I checked the output.
- [X] I applied the [style guide](styleguide.md).

### DIFF
--- a/docs/platform/howto/set-authentication-policies.md
+++ b/docs/platform/howto/set-authentication-policies.md
@@ -4,7 +4,7 @@ title: Set authentication policies for organization users
 
 import ConsoleLabel from "@site/src/components/ConsoleIcons"
 
-The authentication policy for your organization specifies the ways that users in your organization and their personal tokens can access the organization on the Aiven platform.
+The authentication policy for your organization specifies the ways that users in your organization can access the organization on the Aiven Platform.
 
 ## Authentication types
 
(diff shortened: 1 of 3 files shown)


docs/platform/howto/set-authentication-policies.md
```diff
@@ -72,6 +72,13 @@ provider, then the token generated when the user was logged in with their passwo
 not work. After logging in with an allowed method on the new authentication policy
 the user can create a token.
 
+### Access from allowed IP addresses
+
+You can restrict users to accessing the Aiven Platform from specific IP address ranges
```
italomg, line 77: It won't restrict the access to Aiven as a whole, but to the resources belonging to the organization that has IP allow list configured. For example,
User A exists in both org X and Y
if only org X has an IP allow list configured, the user will be required to have an IP in the configured range to access those resources(billing,project,servcie,support,etc..), however the user still has access to the platform and should be able to access resources under org Y for example.

italomg: For managed users this will be different though. Managed users will be effectively locked out of Aiven unless the have an IP within the IP range specified in the allow list of the managing organization

italomg (approved):

wojcik-dorota (approved):
LGTM 🙌🏻

ben-stevens-aiven (approved):
LGTM

PR #613, opened 2024-12-19
fix: links

ArthurFlag:
It seems Docusaurus and Muffet don't get along nicely when a page is in a folder with the same name.


wojcik-dorota (approved):
Thank you 🌷

PR #628, opened 2025-01-10
update(clickhouse): federated queries examples for azureBlobStorage

wojcik-dorota:
https://aiven.atlassian.net/browse/DDB-1580

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/products/clickhouse/howto/run-federated-queries.md
+++ b/docs/products/clickhouse/howto/run-federated-queries.md
@@ -43,7 +43,44 @@ WITH GRANT OPTION allows the user to further transfer the privileges.
 See some examples of running federated queries to read and pull
 data from external S3-compatible object storages.
 
-### Query using SELECT and the S3 function
+### Query using the `azureBlobStorage` table function
+
+#### SELECT and `azureBlobStorage`
+
+```sql
+SELECT *
+FROM azureBlobStorage(
+  'DefaultEndpointsProtocol=https;AccountName=ident;AccountKey=secret;EndpointSuffix=core.windows.net',
+  'ownerresource',
+  'all_stock_data.csv',
+  'CSV',
+  'auto',
+  'Ticker String,
+  Low Float64,
+  High Float64'
+  )
+LIMIT 5
+```
+
+#### INSERT and `azureBlobStorage`
+
+```sql
+INSERT INTO FUNCTION
+  azureBlobStorage(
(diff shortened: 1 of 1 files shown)


docs/products/clickhouse/howto/run-federated-queries.md
```diff
@@ -43,7 +43,44 @@ WITH GRANT OPTION allows the user to further transfer the privileges.
 See some examples of running federated queries to read and pull
 data from external S3-compatible object storages.
 
-### Query using SELECT and the S3 function
+### Query using the `azureBlobStorage` table function
```
tilman-aiven, line 46: For the AzureBlobStorage table engine part

- a named collection is required, no plain connection string with plain credentials are allowed
- therefor we need to reference the "create managed credentials part for azure blob storage"

Create a table
```
CREATE TABLE default.test_azure_table
(
    `Low` Float64,
    `High` Float64
)
ENGINE = AzureBlobStorage(`endpoint_azure-blob-storage-datasets`, blob_path = 'data.csv', compression = 'auto', format = 'CSV')
```
Query from azure blob storage table engine 

```
SELECT avg(Low) FROM test_azure_table
```

tilman-aiven (changes requested):
We should mention "how to get the azure blob storage keys"

You can get the access keys for Azure Blob Storage through the Azure portal, PowerShell, or Azure CLI.

Azure Portal:
```
Go to your storage account in the Azure portal.
Under the resource menu, select "Access keys" to view and copy your account access keys and connection strings.
```

tilman-aiven (changes requested):
thanks! found last minor improvements then we are gtg from my site.

Also is it possible that we mention the source of the data
- in our case https://www.kaggle.com/datasets/jakewright/9000-tickers-of-stock-market-data-full-history?resource=download

something like
```
The query example expects that the file is stored as stocks.csv (link...) in container stockmarket-resource.
```
Then we standardize all queries.

tilman-aiven (approved):
LGTM 👍

docs/products/clickhouse/howto/run-federated-queries.md
```diff
@@ -15,9 +15,12 @@ over an external S3-compatible object storage including relevant S3
 bucket details. A properly constructed federated query returns a
 specific output.
 
-## Before you start
+## Prerequisites
+
+Prerequisites to be met depend on the type of a table function or a table engine to be used
+in your federated query.
```
harshini-rangaswamy, line 21:
```suggestion
The prerequisites depend on the table function or table engine used in your federated query.
```
How about shortening this a bit?

PR #660, opened 2025-01-22
update: manage groups role description

staceysalamon-aiven:
## Checklist

- [ ] I checked the output.


rsalevsky (changes requested):
The change is that group members can't be updated when the group has assigned permissions. (Similar to `organization:app_users:write` `..have not been granted any permissions`)

staceysalamon-aiven:
> The change is that group members can't be updated when the group has assigned permissions. (Similar to `organization:app_users:write` `..have not been granted any permissions`)

Hey @rsalevsky , sorry, with all the different changes in the permissions and conflicts this one got reverted back to the old text accidentally. I've fixed the group description now. Lmk if you want any other changes. Thanks! 🙏

rsalevsky (approved):

ArthurFlag (approved):

PR #688, opened 2025-02-11
feat: Point in time recovery (PITR) for BTAR

wojcik-dorota:
https://aiven.atlassian.net/browse/AD-3009

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/platform/howto/btar/enable-backup-to-another-region.md
+++ b/docs/platform/howto/btar/enable-backup-to-another-region.md
@@ -17,7 +17,11 @@ BTAR is supported for Aiven for MySQL®, Aiven for PostgreSQL®, and Aiven for O
 :::
 
 To add an additional service backup for your service, you can use the Aiven Console, CLI,
-or API.
+or API. You can enable BTAR for
+
+- [New Aiven service](/docs/platform/howto/btar/enable-backup-to-another-region#add-a-cross-region-backup-to-a-new-service)
+  or
+- [Existing Aiven service](/docs/platform/howto/btar/enable-backup-to-another-region#add-a-cross-region-backup-to-an-existing-service)
 
 ## Prerequisites
 
@@ -27,8 +31,8 @@ or API.
 - Depending on your preferred tool to manage BTAR with, make sure you can access or use:
 
   - [Aiven Console](https://console.aiven.io/)
-  - [Aiven CLI client](/docs/tools/cli)
-  - [cURL](https://curl.se/download.html) CLI tool
+  - [Aiven CLI](/docs/tools/cli)
+  - [Aiven API](/docs/tools/api)
 
 ## Add a cross-region backup to an existing service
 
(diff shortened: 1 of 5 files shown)


docs/platform/howto/btar/manage-backup-to-another-region.md
```diff
@@ -73,36 +80,106 @@ As output, you get metrics including replication lags at specific points in time
...
+   1. Set **Backup version** to one of the following:
 
-      :::note
-      For the point-in-time recovery (PITR) option, set up the time to no later than the
-      time of taking the latest backup.
-      :::-->
+      - **Latest transaction**
+      - **Point in time**: Set it up to no earlier than the time of taking the oldest
+        replicated base backup and no later than the time of taking the latest replicated
+        base backup.
```
ali-gholami-aiven, line 109: actually `no later than the time of taking the latest replicated base backup` is not true. users can set recovery target time **after the latest replicated base backup** as well. the time range end gets updated frequently, every time we replicate a new WAL (for PG) or Binlog (for MySQL) file.

So for example if service is started on 2025-02-10, BTAR is enabled immediately so we have 3 base backups:

bb1: `2025-02-10 10:00:00Z`
bb2: `2025-02-11 10:00:00Z`
bb3: `2025-02-12 10:00:00Z`

and current time is `2025-02-12 18:00:00Z`, then the time range would be 
From `2025-02-10 10:00:00`  To (approx) `2025-02-10 17:55:00`.

But now I'm thinking this end range is not visible to users, although when they insert it we do the validation and we won't let them to set a time after the `2025-02-10 17:55:00`

I'm not quite sure how to explain this without getting into implementation details, though.

ali-gholami-aiven (changes requested):

ali-gholami-aiven (approved):
LGTM

docs/platform/howto/btar/manage-backup-to-another-region.md
bugant: @ali-gholami-aiven @wojcik-dorota this looks still wrong to me. With PITR we aim at allowing the customer to be able to fork and get the sate of at most 1h ago (IIRC).

bugant (changes requested):

docs/platform/howto/btar/enable-backup-to-another-region.md
```diff
@@ -46,21 +56,68 @@ To add an additional service backup for your service, you can use the Aiven
...
-:::note[additional_backup_regions]
-To enable BTAR on an Aiven service, add the `additional_backup_regions`
-parameter to relevant commands.
-:::
+## Add a cross-region backup to a new service
+
+<Tabs groupId="group1">
+<TabItem value="gui" label="Aiven Console" default>
+
+#### 1. Create a service
```
ArthurFlag, line 98: If I'm not mistaken, most of the docs use the `Step X:` prefix for such cases.
 
```suggestion
#### Step 1: Create a service
```

docs/platform/howto/btar/enable-backup-to-another-region.md
```diff
@@ -46,21 +56,68 @@ To add an additional service backup for your service, you can use the Aiven
...
 
-### Create a service with BTAR via CLI {#new-service-cli}
+<CreateService/>
+
+#### 2. Add a backup for the new service
```
ArthurFlag, line 102:
```suggestion
#### Step 2: Add a backup for the new service
```

ArthurFlag (approved):

PR #740, opened 2025-03-04
update: list of alloydb omni extensions

wojcik-dorota:
https://aiven.atlassian.net/browse/BF-3137

## Checklist

- [x] I checked the output.


rdunklau:
Should we instead refer to google's own documentation and only list the differences ? 
I'm not sure how much value there is in maintaining this when Google already does.

docs/products/alloydbomni/reference/list-of-extensions.md
```diff
@@ -0,0 +1,189 @@
...
+
+To list the extensions and their details, such as extension version numbers,
+connect to your service and run:
+
+```sql
+SELECT *
+FROM pg_available_extensions;
+```
+
+## Limitations
```
rdunklau, line 19: This doesn't apply to Google's Alloydb-omni, as we give the alloydbadmin role to our users which allow them to create extensions normally restricted to superusers.

docs/products/alloydbomni/reference/list-of-extensions.md
```diff
@@ -0,0 +1,189 @@
...
+-   [ip4r](https://github.com/RhodiumToad/ip4r): Provides data types for IPv4/v6 addresses,
+    and address ranges, plus index support
+-   [isn](https://www.postgresql.org/docs/current/isn.html): Data types
+    for international product numbering standards
+-   [ltree](https://www.postgresql.org/docs/current/ltree.html): Data
+    type for hierarchical tree-like structures
+-   [pgvector](https://github.com/pgvector/pgvector): Type for vector
+    similarity search; `PG13 and newer`
+-   [postgresql-hll](https://github.com/citusdata/postgresql-hll): Type for storing
+    `hyperloglog` data; `PG11 and newer`
```
rdunklau, line 63: AlloyDB is currently 15 only, and will therefore never have a version inferior to 15 so I would remove all references to the versioning number.

docs/products/alloydbomni/reference/list-of-extensions.md
```diff
@@ -0,0 +1,189 @@
...
+  Functions for tracking who changed a table
+- [intagg](https://www.postgresql.org/docs/current/intagg.html).
+  Integer aggregator and enumerator (obsolete).
+- [intarray](https://www.postgresql.org/docs/current/intarray.html):
+  Functions, operators, and index support for 1-D arrays of integers
+- [lo](https://www.postgresql.org/docs/current/lo.html): Large Object
+  maintenance
+- [moddatetime](https://www.postgresql.org/docs/current/contrib-spi.html#CONTRIB-SPI-MODDATETIME):
+  Functions for tracking last modification time
+- [pageinspect](https://www.postgresql.org/docs/current/pageinspect.html):
```
rdunklau, line 144: Some of those have been removed by us, see ticket here for the list: 
https://github.com/aiven/aiven-core/pull/51357

rdunklau (changes requested):

wojcik-dorota:
> Should we instead refer to google's own documentation and only list the differences ? I'm not sure how much value there is in maintaining this when Google already does.

Thanks, Ronan :-) Discussing that with Arthur and it seems it'd be still easier to maintain the list if we include it directly in our doc.

rdunklau (approved):

PR #747, opened 2025-03-10
update: add note about Flink service creation limitation

harshini-rangaswamy:
## Describe your changes

Added note about Flink service creation limitation. 

Preview - https://hh-5190-document-new-flink-c.aiven-docs.pages.dev/docs/products/flink
[HH-5190](https://aiven.atlassian.net/browse/HH-5190)

## Checklist

- [x] I checked the output.

[HH-5190]: https://aiven.atlassian.net/browse/HH-5190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/includes/flink-service-creation-limitation.md
+++ b/static/includes/flink-service-creation-limitation.md
@@ -3,8 +3,8 @@
 **Aiven for Apache Flink® service creation is limited**
 
 Creating a new Aiven for Apache Flink® service is currently limited.
-If your organization does not have access to create an Aiven for Apache service, open a
-[support ticket](/docs/platform/howto/support#create-a-support-ticket) or
+If your organization does not have access to create an Aiven for Apache Flink service,
+open a [support ticket](/docs/platform/howto/support#create-a-support-ticket) or
 contact [Aiven Support](mailto:support@aiven.io) for assistance.
 
 :::


static/includes/flink-service-creation-limitation.md
```diff
@@ -0,0 +1,11 @@
+:::note
+
+**Aiven for Apache Flink® service creation is limited**
+
+Creating an Aiven for Apache Flink service is limited to existing accounts. If you do not
```
staceysalamon-aiven, line 5: I'm not sure I understand the note - is Flink only available for users who signed up for an Aiven account prior to a specific date? Because it says "if you do not have an account" it sounds to me like I can just sign up and then create a service.

staceysalamon-aiven (commented):
Just have a question about the specifics of the limitation.

static/includes/flink-service-creation-limitation.md
harshini-rangaswamy: The point is that only existing Aiven accounts can create an Aiven for Apache Flink service at this time. New users cannot sign up and create a Flink service directly. I don't believe there is a specific date, but it would be good to confirm. @MichaelTansiniAiven , is there a specific date for this?

staceysalamon-aiven: Okay, I thought that might be the case, but a user may not know whether their account falls into the category of "new". That could be clarified and also the next steps sound a little confusing to me since we're telling the users to open a ticket in Aiven Console, which requires an account.

I'm not sure how exactly to reword it, however. :)

harshini-rangaswamy: That's a good point. @MichaelTansiniAiven, is there a backend check that determines whether an account qualifies as 'existing'? I assume an error would be displayed if a user with a new account tries to create a Flink service. How is this restriction enforced on the backend? If a backend check is not in place, then I think we should hold off on publishing this note. 

>  the next steps sound a little confusing to me since we're telling the users to open a ticket in Aiven Console, which requires an account.

Is an account tied to a specific service or the platform as a whole? Or maybe 'account' isn't the right term here? I didn't find much information on what 'account' means in the docs.

staceysalamon-aiven: The word `account` has some baggage at Aiven and we've done a bit of work in the UI and most of the platform docs to remove this word since it basically used to refer to what are now orgs and units and that could cause confusion. My personal approach has been to use the phrase "user account" for an individual's account on Aiven.

But this is a good question because maybe access to Flink is based on the org?

static/includes/flink-service-creation-limitation.md
```diff
@@ -0,0 +1,10 @@
+:::note
+
+**Aiven for Apache Flink® service creation is limited**
+
+Creating a new Aiven for Apache Flink® service is currently limited.
+If your organization does not have access to create an Aiven for Apache service, open a
```
staceysalamon-aiven, line 6: Totally missed this on the first review. 🙃

```suggestion
If your organization does not have access to create an Aiven for Apache Flink service, open a
```

staceysalamon-aiven (approved):
LGTM! I just now noticed that the full product name was missing in one part (didn't see it on my first review).

PR #757, opened 2025-03-13
cassandra: update EOL and EOA

alanfranz:
Update Aiven for Cassandra EOL and EOA according to our previous customer messaging.


wojcik-dorota (approved):
Thank you 🥇

PR #767, opened 2025-03-18
update: pg replication slots update

wojcik-dorota:
https://aiven.atlassian.net/browse/BF-3193

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/postgresql/concepts/upgrade-failover.md
+++ b/docs/products/postgresql/concepts/upgrade-failover.md
@@ -135,11 +135,11 @@ post](https://aiven.io/blog/aiven-for-pg-recreates-logical-replication-slots).
 :::important
 
 - Replication slots are not recovered after major version upgrades of Aiven for PostgreSQL.
-- To prevent losing a replication slot when promoting a new primary server, all changes
-  from the old primary have to be fully replicated before promotion. This requires waiting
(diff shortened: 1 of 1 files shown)


docs/products/postgresql/concepts/upgrade-failover.md
```diff
@@ -133,8 +133,14 @@ post](https://aiven.io/blog/aiven-for-pg-recreates-logical-replication-slots).
 :::
 
 :::important
-Replication slots are not recovered after major version upgrades of
-Aiven for PostgreSQL.
+
+- Replication slots are not recovered after major version upgrades of Aiven for PostgreSQL.
+- To prevent losing a replication slot when promoting a new primary server, all changes
+  from the old primary have to be fully replicated before promotion. This requires waiting
```
rdunklau, line 139: This is actually not about changes being replicated, but having the history.
I would rephrase this as: 
- To prevent losing a replication slot when promoting a new primary server, all changes
  from the old primary have to be consumed so that it's catalog_xmin does not predate the standby server. If after 30 minutes, the slot has still not been consumed the syncing progress will still be marked as done and there is a risk that the logical slot will be lost upon promotion of the new node.

palmieric (approved):
LGTM!

staceysalamon-aiven (approved):

PR #779, opened 2025-03-28
fix: Kafka services do not have replica uri

tvainika:
## Describe your changes

Prometheus client needs to have additional replica uri if multi node service has separate replica uri. Kafka is multi node service, but it does not have replica uri. For Kafka all nodes are behind one uri.

## Checklist

- [X] I checked the output.
- [X] I applied the [style guide](styleguide.md).


harshini-rangaswamy (approved):
Thanks for updating this!

PR #781, opened 2025-03-28
update: max number of ClickHouse DBs

wojcik-dorota:
https://aiven.atlassian.net/browse/DDB-1834

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/clickhouse/reference/limitations.md
+++ b/docs/products/clickhouse/reference/limitations.md
@@ -143,6 +143,15 @@ a
         or the [Aiven Console](https://console.aiven.io/).
       </td>
     </tr>
+    <tr>
+      <td>Maximum number of databases per service</td>
+      <td>
+          Your Aiven for ClickHouse service can support up to 400 databases simultaneously.
+      </td>
+      <td>
+        N/A
+      </td>
+    </tr>
   </tbody>
 </table>
 
--- a/docs/products/clickhouse/reference/limitations.md
+++ b/docs/products/clickhouse/reference/limitations.md
@@ -149,7 +149,11 @@ a
           Your Aiven for ClickHouse service can support up to 400 databases simultaneously.
       </td>
       <td>
-        N/A
+        Instead of creating multiple databases of the same structure for isolation purposes,
+        it's recommended to create one database where you add an extra column to filter by
+        it. Consider including this column into the primary key or partitioning by it. You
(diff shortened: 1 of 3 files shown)


docs/products/clickhouse/reference/limitations.md
```diff
@@ -143,6 +143,15 @@ a
         or the [Aiven Console](https://console.aiven.io/).
       </td>
     </tr>
+    <tr>
+      <td>Maximum number of databases per service</td>
+      <td>
+          Your Aiven for ClickHouse service can support up to 400 databases simultaneously.
+      </td>
+      <td>
+        N/A
```
Khatskevich, line 152: Everything looks good.
Only in this place we could add some recommendations.
I do not know how to formulate it properly, but the main mistake we saw was:
Our client has many customers, and for some reason he wants to store data of each customer in a separate database. And he creates the same database with the same tables for each customer.
It leads to small tables and parts and to data being distributed between many small files on the disk. It creates redundant extra load and slows down insertion and selection.

We would recommend in this case to store all the clients together by adding a "client" column to the table (and possibly including the column into the primary key).
Even in case our client gives his customers ability to run SQL against their own data, our client, they could limit their scope using `additional_table_filters` query setting to limit their scope.

docs/products/clickhouse/reference/limitations.md
```diff
@@ -143,6 +143,19 @@ a
...
+      <td>
+          Your Aiven for ClickHouse service can support up to 400 databases simultaneously.
+      </td>
+      <td>
+        If your use case requires more than 400 databases, it's recommended to consolidate
+        data by adding a `client` column to your tables and including the column in the
+        primary key. This approach allows you to store all client data together while
+        maintaining separation. You can also limit scope of data on which SQL queries can
+        be run by using the `additional_table_filters` query setting.
+      </td>
```
Khatskevich, line 157: I suppose it might be misunderstood. Possible change:
```
If you are considering creating many databases of the same structure for isolation purposes, it's 
recommended to add extra column and filter by it instead. It will allow better scalability, and improve
robustness. A good idea would be to include this column into primary key or even partition by it.
You can also limit scope of data on which SQL queries can be run by using the `additional_table_filters`
query setting.
```

Khatskevich (approved):

harshini-rangaswamy (approved):
LGTM!

PR #790, opened 2025-04-03
update: Kafka and Kafka Connect get started Terraform instructions

staceysalamon-aiven:
## Describe your changes

Replaces TF code with synced samples from the TF repository. Adds TF code to the Kafka Connect get started guide.

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/products/kafka/get-started.md
+++ b/docs/products/kafka/get-started.md
@@ -59,7 +59,7 @@ or automate the process with Terraform.
 
 1. Create a file named `service.tf` and add the following:
 
-    <TerraformSample filename='kafka_connect/service.tf' />
+    <TerraformSample filename='kafka_connect/kafka_service.tf' />
 
 1. Create a file named `variables.tf` and add the following:
 
--- a/docs/products/kafka/get-started.md
+++ b/docs/products/kafka/get-started.md
@@ -27,12 +27,12 @@ Before you begin, ensure you have access to the following tools:
 - Access to the [Aiven Console](https://console.aiven.io)
 
 </TabItem>
-<TabItem value="terraform" label="Terraform" default>
+<TabItem value="terraform" label="Terraform">
 
 <TerraformPrereqs />
 
 </TabItem>
-<TabItem value="cli" label="CLI" default>
+<TabItem value="cli" label="CLI">
 
 - [Aiven CLI](https://github.com/aiven/aiven-client#installation) installed
 - [A personal token](https://docs.aiven.io/docs/platform/howto/create_authentication_token.html)
--- a/docs/products/kafka/get-started.md
+++ b/docs/products/kafka/get-started.md
@@ -10,7 +10,10 @@ import ConsoleLabel from "@site/src/components/ConsoleIcons"
 import CreateService from "@site/static/includes/create-service-console.md"
 import LimitedBadge from "@site/src/components/Badges/LimitedBadge";
 import EarlyBadge from "@site/src/components/Badges/EarlyBadge";
-import RelatedPages from "@site/src/components/RelatedPages"
+import RelatedPages from "@site/src/components/RelatedPages";
+import TerraformPrereqs from "@site/static/includes/terraform-get-started-prerequisites.md";
+import TerraformApply from "@site/static/includes/terraform-apply-changes.md";
+import TerraformSample from '@site/src/components/CodeSamples/TerraformSample';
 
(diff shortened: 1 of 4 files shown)


docs/products/kafka/get-started.md
```diff
@@ -10,114 +10,76 @@ import ConsoleLabel from "@site/src/components/ConsoleIcons"
...
-
-   output "kafka_service_password" {
-     value     = aiven_kafka.kafka.service_password
-     sensitive = true
-   }
-   ```
+<TabItem value="terraform" label="Terraform">
 
-1. Create a `terraform.tfvars` file to assign values to your declared variables.
+The following example files are part of the Kafka Connect example in the
```
harshini-rangaswamy, line 56: I think it would be better to keep this section focused just on creating the Kafka service via Terraform. We could probably link to this instead:
https://github.com/aiven/terraform-provider-aiven/blob/main/docs/resources/kafka.md

The Kafka Connect part could be moved either to the Connect to your Aiven for Apache Kafka service section (#connect) or to the Kafka Connect getting started doc:
docs/products/kafka/kafka-connect/get-started.md

harshini-rangaswamy: Yes, I totally agree with avoiding duplication and pointing to the full example.

What I meant was that for the Create Kafka service via Terraform section, we should use an example that focuses just on creating the Kafka service rather than using the Kafka Connect example. I found this in the [TF dcos](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/kafka), which I thought might fit better here. As you pointed out, the Kafka Connect content has already been added to this PR.

staceysalamon-aiven: So, this example does focus on only Kafka. If you look at the preview, you'll see that the files are just for that and if you run them all together that's all that gets created. The example usage that's in the TF docs is basically the same as what is in the step where you create a file called `service.tf`. 

I can remove the link to the GitHub repo or reword this intro if you think it is confusing, but the files in this guide only create a Kafka service.

staceysalamon-aiven: This is the section I mean: 
{E138B163-3B06-4C0B-BF27-98491E4CEF7B}

staceysalamon-aiven: Different files are inserted for the Kafka Connect get started guide so that you are only creating a Connect service and integrating with the existing Kafka (as mentioned in the prereqs section): 
{AF4376C1-667C-4CAD-9403-630355EBA486}

harshini-rangaswamy: image
image


I get that different files are inserted for the Kafka Connect getting started guide, but my concern is more around the line in the **Kafka get started** section:  
**"The following example files are part of the Kafka Connect example..."**

This example only creates a Kafka service, but we say it's part of a combined Kafka + Kafka Connect setup. That could be a bit confusing, since some variables (like `kafka_connect_name`) are present even though they're not used when you first create a Kafka service. It works fine for creating Kafka, but it might not be the cleanest or most minimal example if the goal is to show just Kafka service creation.

Whereas in the [Terraform docs](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/kafka), the example focuses solely on creating and managing a Kafka service—which might be a better fit here instead of one that includes Kafka Connect.

staceysalamon-aiven: But we can't use the example usage alone here. For get started guides we have to have the full set of files. I can refine the introduction to make it clearer that this only creates a Kafka service.

harshini-rangaswamy: > make it clearer that this only creates a Kafka service.

Sounds good, thanks!

staceysalamon-aiven: Restructured it a little anyway and linked to the connect guide for clarity.

harshini-rangaswamy: This should work for now. Just out of curiosity, what's the reasoning behind including the Kafka service within the Kafka Connect example, as seen [here](https://github.com/aiven/terraform-provider-aiven/tree/main/examples/kafka_connect)?

Users can either create a dedicated Kafka Connect service or integrate it with an existing Kafka service. Kafka Connect is not required to use Kafka—it's mainly used when you need to stream data between Kafka and external systems.  Users can also use a Kafka service without Kafka Connect.

harshini-rangaswamy (approved):
Thanks for this, looks good!

PR #819, opened 2025-04-16
update: use external identity example from TF repo

harshini-rangaswamy:
## Describe your changes
Update docs to use the external identity example from the Terraform repo. 

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/products/kafka/howto/terraform-governance-approvals.md
+++ b/docs/products/kafka/howto/terraform-governance-approvals.md
@@ -95,35 +95,12 @@ be integrated to validate changes and enforce compliance.
 ### Step 2. Map GitHub users to Aiven identities
 
 To verify requesters and approvers, assign them to the appropriate user group
-using the `aiven_external_identity` resource. For more information, see the
+using the `aiven_external_identity` resource.
+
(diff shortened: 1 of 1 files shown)


docs/products/kafka/howto/terraform-governance-approvals.md
```diff
@@ -95,35 +95,12 @@ be integrated to validate changes and enforce compliance.
 ### Step 2. Map GitHub users to Aiven identities
 
 To verify requesters and approvers, assign them to the appropriate user group
```
staceysalamon-aiven, line 97: This is not something I'm super familiar with, but should this instead say something about the user ID mapping instead of groups since it's now only the external_identity resource?
```suggestion
To verify requesters and approvers, map their GitHub user IDs to their Aiven user IDs
```

staceysalamon-aiven (approved):
Embedded sample looks great! Just one question about the intro to this step.

docs/products/kafka/howto/terraform-governance-approvals.md
harshini-rangaswamy: Good point and makes sense. I will update this.

PR #832, opened 2025-04-27
feat: manage custom repos and snapshots via console

wojcik-dorota:
https://aiven.atlassian.net/browse/MA-3312

## Checklist

- [ ] I checked the output.


src/components/ConsoleIcons/index.tsx
```diff
@@ -603,6 +627,13 @@ export default function ConsoleLabel({name}): ReactElement {
           <b>Replication flow</b>
         </>
       );
+    case 'snapshots':
+      return (
+        <>
+          <ConsoleIconWrapper icon={ConsoleIcons.moveProject} />{' '}
+          <b>Snapshots</b>
```
nctbao, line 634: Discussed with @mteiste and @axeoman and we thought that `Snapshots` is confusing because it's essentially the same with Backups in OS context. Suggested to go with `Custom snapshots` to be more explicit about this. What do you think?

nctbao (commented):
This looks really good 💎  I noticed few small differences from Console below and also added a comment about the menu text after our discussion today with Marko and Aleksei.

src/components/ConsoleIcons/index.tsx
wojcik-dorota: @nctbao , I totally get the need to differentiate btw backup & snapshot but adding "custom" doesn't help me understand it to be honest and makes me wonder: why "custom"? in what sense?
Although it's not a perfect solution, I'd stick to "Snapshots" expecting most of our users to be more or less familiar with OpenSearch functionalities and terminology.
Ofc, we can still use "custom snapshots", likely it's just me not seeing value added here :-)

nctbao (approved):
Looks great! 💎

docs/products/opensearch/howto/manage-snapshots.md
```diff
@@ -0,0 +1,402 @@
...
+
+   :::tip
+   If your click **Close** in the **Restore snapshot** window during the restore process,
+   later you can check the restore process status on the **Snapshots** page.
+   :::
+
+1. When the **Restore snapshot** window shows that the restore process is completed,
+   click **Close**.
+
+</TabItem>
```
nctbao, line 178: @wojcik-dorota sorry, I just noticed that restore snapshot is not yet supported in Aiven Console. Could we hide this until it's available? :)

wojcik-dorota: I'll merge the PR when the console is supported :-)

PR #1161, opened 2025-11-01
update: how to delete PG extensions

wojcik-dorota:
https://aiven.atlassian.net/browse/BF-3742

[DOC PREVIEW](https://bf-3742-pg-extensions-can-be.aiven-docs.pages.dev/docs/products/postgresql/howto/manage-extensions#delete-an-extension)

## Checklist

- [x] I checked the output.


harshini-rangaswamy (approved):
LGTM. One small thought to consider: do we want to call out explicitly that any service user can drop extensions, per the bug note?

Feel free to leave as is if you think it’s already clear.

kathia-barahona (commented):
lgtm

PR #1326, opened 2026-02-12
feat: controlled upgrade in LA

wojcik-dorota:
https://aiven.atlassian.net/browse/IP-774

[DOC PREVIEW](https://ip-774-docs-public-documenta.aiven-docs.pages.dev/docs/platform/howto/controlled-upgrade)

Followed up on in https://github.com/aiven/aiven-docs-test/pull/180

### DIFF
--- a/docs/platform/howto/controlled-upgrade.md
+++ b/docs/platform/howto/controlled-upgrade.md
@@ -8,6 +8,13 @@ import RelatedPages from "@site/src/components/RelatedPages";
 
 Ensure your source service is updated and validated before the same maintenance update reaches your destination service.
 
+:::important
+Controlled upgrade pipeline is a
+[limited availability](/docs/platform/concepts/service-and-feature-releases#limited-availability-)
+feature.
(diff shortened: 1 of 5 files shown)


docs/platform/howto/controlled-upgrade.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Use controlled upgrade pipelines
+
+### Prerequisites
+
+To use controlled upgrade pipelines:
+
+- [Aiven CLI](/docs/tools/cli) or [Aiven API](/docs/tools/api)
+- Write access to the source and destination projects
+- The correct access control list (ACL) permissions for the feature in your destination
```
heikju: I think about this it would be enough to communicate that this needs to be enabled by Aiven, like any other Limited Availability feature. Referring to our internal mechanism of doing so doesn't seem very useful here either.

heikju (dismissed):
There are some references to internal implementation details of ours that don't seem very useful to a customer, and since the feature is incomplete and no client support exists yet we'll likely want to wait with this for a bit.

PR #1504, opened 2026-05-04
update: reorganize OS doc IA

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-1762

[DOC PREVIEW](https://doc-1762-replace-diataxis-wi.aiven-docs.pages.dev/docs/products/opensearch)

### DIFF
--- a/.github/vale/styles/config/vocabularies/Aiven/accept.txt
+++ b/.github/vale/styles/config/vocabularies/Aiven/accept.txt
@@ -322,6 +322,8 @@ reindexed
 reindexing
 remoteSecure
 renewables
+repo
+repos
 requesters
 resharding
 Resharding
--- a/docs/products/opensearch/concepts/access_control.md
+++ b/docs/products/opensearch/concepts/access_control.md
@@ -1,6 +1,6 @@
 ---
 title: Access control in Aiven for OpenSearch®
-sidebar_label: Access control
+sidebar_label: Control access to services
 ---
 
 Access control is a crucial security measure that allows you to control who can access your data and resources. By setting up access control rules, you can restrict access to sensitive data and prevent unauthorized changes or deletions.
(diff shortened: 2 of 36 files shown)


andreitrepet:
Hi @wojcik-dorota,
Thanks for putting this together! I've reviewed the new structure, and here are a few suggestions:
- Move **Cluster architecture and replication** to position 3 (after **Connect to OS services**)
- Make **OpenSearch Dashboards** a standalone section
- Rename **Searching & aggregation** to **Searching and aggregating data** and make it a standalone section
- Rename ‘Observability & monitoring’ to ‘Integrations’
- Move 'Security & access' to position after 'Querying data’
- **Optimal shards number** article
    - Move to **Indexing** (put it under Indices on position 3 (after **When to create an index**)
    - Link the **Setting the number and size of shards sub-section** in the Indices article to the **Optimal shards number** article)
- **Manage large shards in OpenSearch** article - move to the **Indexing** section
- ‘Cluster architecture and replication section’ - article order:
    - High availability
    - Dedicated node roles
    - Cross-cluster replication

UX-Irene:
Hi, I can see the order is a bit different and I get it with Opensearch it makes more sense to have the manage service and manage cluster  and then the  Index and storage , OpenSearch® Dashboards
 and Searching and aggregation. So let leave this as you have it here and revisit for the other services, perhaps we move up manage service later, to be more like the Opensearch version. 

What I would recommend is to rearrange the last ones to be consistent  with the others. Unless you feel strongly about your order?

Backups and migration
Integrations & extensions
Security and access

wojcik-dorota:
> Hi, I can see the order is a bit different and I get it with Opensearch it makes more sense to have the manage service and manage cluster and then the Index and storage , OpenSearch® Dashboards and Searching and aggregation. So let leave this as you have it here and revisit for the other services, perhaps we move up manage service later, to be more like the Opensearch version.
> 
> What I would recommend is to rearrange the last ones to be consistent with the others. Unless you feel strongly about your order?
> 
> Backups and migration Integrations & extensions Security and access

Sure thing, done! Thanks 🙏🏻

UX-Irene (approved):
Restructure looks good and makes sense. Noticed you moved Manage Service higher up, so this is the only thing we need to align on with the other service. We will tackle these high-level placement at the end.

PR #1518, opened 2026-05-08
fix: Improve agent-friendly docs support and llms.txt discovery

harshini-rangaswamy:
## Describe your changes

This PR improves agent-friendly access to Aiven documentation by making `llms.txt` easier to discover, keeping it small enough for agents to read, and adding discovery hints to generated markdown files.

The changes target documentation-side AI readiness. Remaining issues such as `Accept: text/markdown` content negotiation and soft 404 behavior require Cloudflare/hosting-layer changes.

### Added `llms.txt` discovery in HTML

Updated `docusaurus.config.ts` to add a site-wide alternate link in the page `<head>`:

```html
<link rel="alternate" type="text/plain" href="/docs/llms.txt">
```

The URL is derived from `baseUrl`, so it works across local, preview, and production environments.

### Added hidden page-level discovery directives

Added hidden agent-readable directives to:

```
src/pages/index.tsx
src/theme/DocRoot/index.tsx
```

These directives help agents that inspect page body content find `llms.txt`.

The implementation:

- uses `useBaseUrl('llms.txt')` to make the link base URL-aware across all environments
- avoids hardcoded `/docs/` paths that would break in preview or staging deployments
- uses `aria-hidden="true"` so screen readers do not read hidden agent-only content
- uses `tabIndex={-1}` to remove the hidden link from keyboard navigation
- relies on `.agent-docs-directive` in `custom.css` for visual hiding, avoiding duplicate inline styles

### Added markdown discovery directives

Added `scripts/add-markdown-agent-directive.mjs`.

This post-build script prepends an `llms.txt` pointer to generated `.md` files:

```md
> For the complete documentation index, see [llms.txt](/docs/llms.txt).
```

The link is derived from `BASEURL`, and the script logs how many markdown files were modified out of the total files found.

### Trimmed `llms.txt`

Added `scripts/trim-llms-txt.mjs`.

This script runs after the Docusaurus build and trims `build/llms.txt` below the configured 48,000-character cap. The previous file was 273K — most agents truncate above 100K, making the majority of the docs invisible.

It also appends a footer linking to the full index:

```md
## Full documentation index

- https://aiven.io/docs/llms-full.txt
```

The footer URL is derived from `BASEURL`, and the script reserves footer space before trimming so the final file stays under the cap.

### Updated build pipeline

Updated `package.json` so the standard build runs the post-build processing automatically:

```json
"build": "docusaurus build && node scripts/trim-llms-txt.mjs && node scripts/add-markdown-agent-directive.mjs"
```

No manual step is required.

## Why this matters

AI agents and tools often read documentation differently from humans. They may look for `llms.txt`, parse raw markdown, or inspect page metadata instead of using the browser UI.

Before these changes:

- `llms.txt` was 273K and likely to be truncated by most agents
- HTML pages did not point agents to `llms.txt`
- Markdown pages did not point agents to the full documentation index
- Some links were at risk of being inconsistent across base URL environments

After these changes:

- `llms.txt` is under 48K and fully readable by agents
- `llms-full.txt` remains available for the complete index
- HTML and markdown pages both point agents to `llms.txt`
- All links are base URL-aware and work across local, preview, and production
- Hidden directives do not affect the visible UI or keyboard navigation

### 1. Build the site

```bash
yarn build
```

Confirm the post-build scripts run. The terminal output should include:

```
Trimmed build/llms.txt to 47893 characters.
Added directive to 866/866 markdown files.
```

### 2. Start the local server

```bash
yarn serve
```

Note the port shown in the terminal. The examples below use `3000`.

### 3. Check `llms.txt`

Verify the file is under the 48,000-character cap:

```bash
wc -c build/llms.txt
```

Expected:

```
< 48000 build/llms.txt
```

Verify the footer links to the full index:

```bash
grep 'llms-full.txt' build/llms.txt
```

Expected:

```
- https://aiven.io/docs/llms-full.txt
```

### 4. Check `llms-full.txt` is reachable

```bash
curl -I http://localhost:3000/docs/llms-full.txt
```

Expected:

```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
```

### 5. Check HTML discovery

Verify the `<head>` link and hidden directive are present:

```bash
grep -R 'llms.txt' build/index.html | head
```

Expected output includes:

```html
<link rel="alternate" type="text/plain" href="/docs/llms.txt">
```

And the hidden directive text:

```
For the complete documentation index, see
```

### 6. Check markdown discovery

Verify a doc page has the directive prepended:

```bash
head -5 build/docs/get-started.md
```

Expected:

```
> For the complete documentation index, see [llms.txt](/docs/llms.txt).
```

### 7. Run afdocs

```bash
npx afdocs check http://localhost:3000/docs/ --fixes --verbose
```

Expected result:

```
16 passed, 1 warning, 2 failed, 4 skipped
```

Expected passing checks:

```
llms-txt-exists
llms-txt-valid
llms-txt-size
llms-txt-directive-html
llms-txt-directive-md
markdown-url-support
markdown-content-parity
```

Expected failures (out of scope for this PR — require Cloudflare/hosting changes):

```
content-negotiation
http-status-codes
```

## Validation

Local production build was tested with `afdocs`.

Result:

```
16 passed
1 warning
2 failed
4 skipped
```

Generated file checks:

```
build/llms.txt: 47,893 characters
footer link: https://aiven.io/docs/llms-full.txt
```

## Remaining follow-up work

These are not addressed in this PR because they require routing or hosting changes.

### Content negotiation

The server still returns HTML for requests with:

```
Accept: text/markdown
```

This likely needs Cloudflare configuration or a Worker.

### Soft 404 handling

The local static server returns `200` for nonexistent URLs. This should be verified in production or handled at the hosting/CDN layer.

### Content start position warning

One local warning remains because the docs landing page content starts after some layout/navigation markup. This is not related to the `llms.txt` work and is not a blocker for this PR.

## Checklist

- [x] I checked the output.

### DIFF
--- a/scripts/add-markdown-agent-directive.mjs
+++ b/scripts/add-markdown-agent-directive.mjs
@@ -24,15 +24,15 @@ async function findMarkdownFiles(dir) {
 }
 
 const files = await findMarkdownFiles(BUILD_DIR);
+let modified = 0;
 
 for (const file of files) {
   const content = await fs.readFile(file, 'utf8');
 
   if (!content.startsWith(DIRECTIVE)) {
     await fs.writeFile(file, DIRECTIVE + content);
+    modified++;
   }
 }
 
-console.log(
-  `Added markdown agent directive to ${files.length} markdown files.`,
-);
+console.log(`Added directive to ${modified}/${files.length} markdown files.`);
--- a/scripts/add-markdown-agent-directive.mjs
+++ b/scripts/add-markdown-agent-directive.mjs
@@ -7,7 +7,13 @@ const llmsTxtUrl = `${baseUrl.replace(/\/$/, '')}/llms.txt`;
 const DIRECTIVE = `> For the complete documentation index, see [llms.txt](${llmsTxtUrl}).\n\n`;
 
 async function findMarkdownFiles(dir) {
-  const entries = await fs.readdir(dir, {withFileTypes: true});
+  let entries;
+  try {
+    entries = await fs.readdir(dir, {withFileTypes: true});
+  } catch (err) {
+    throw new Error(`Failed to read directory ${dir}: ${err.message}`, {cause: err});
+  }
+
   const files = [];
 
   for (const entry of entries) {
@@ -23,16 +29,23 @@ async function findMarkdownFiles(dir) {
   return files;
 }
 
-const files = await findMarkdownFiles(BUILD_DIR);
-let modified = 0;
+async function main() {
+  const files = await findMarkdownFiles(BUILD_DIR);
+  let modified = 0;
 
-for (const file of files) {
-  const content = await fs.readFile(file, 'utf8');
+  for (const file of files) {
+    const content = await fs.readFile(file, 'utf8');
 
-  if (!content.startsWith(DIRECTIVE)) {
-    await fs.writeFile(file, DIRECTIVE + content);
-    modified++;
+    if (!content.startsWith(DIRECTIVE)) {
+      await fs.writeFile(file, DIRECTIVE + content);
+      modified++;
+    }
   }
+
+  console.log(`Added directive to ${modified}/${files.length} markdown files.`);
 }
 
-console.log(`Added directive to ${modified}/${files.length} markdown files.`);
+main().catch((err) => {
+  console.error(err);
+  process.exit(1);
+});
--- a/scripts/add-markdown-agent-directive.mjs
+++ b/scripts/add-markdown-agent-directive.mjs
@@ -1,0 +1,38 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+
+const BUILD_DIR = 'build';
+const baseUrl = process.env.BASEURL || '/docs/';
+const llmsTxtUrl = `${baseUrl.replace(/\/$/, '')}/llms.txt`;
+const DIRECTIVE = `> For the complete documentation index, see [llms.txt](${llmsTxtUrl}).\n\n`;
+
+async function findMarkdownFiles(dir) {
+  const entries = await fs.readdir(dir, {withFileTypes: true});
+  const files = [];
+
+  for (const entry of entries) {
+    const fullPath = path.join(dir, entry.name);
+
+    if (entry.isDirectory()) {
+      files.push(...(await findMarkdownFiles(fullPath)));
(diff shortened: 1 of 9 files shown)


scripts/add-markdown-agent-directive.mjs
```diff
@@ -0,0 +1,38 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+
+const BUILD_DIR = 'build';
+const baseUrl = process.env.BASEURL || '/docs/';
+const llmsTxtUrl = `${baseUrl.replace(/\/$/, '')}/llms.txt`;
+const DIRECTIVE = `> For the complete documentation index, see [llms.txt](${llmsTxtUrl}).\n\n`;
+
+async function findMarkdownFiles(dir) {
+  const entries = await fs.readdir(dir, {withFileTypes: true});
```
banafshr: since fs.readdir is a promise, it's always good to add it in try/error. but it's not a deal breaker since this only runs in the CI

scripts/limit-llms-txt-size.mjs
```diff
@@ -0,0 +1,41 @@
...
+  `- ${FULL_INDEX_URL}`,
+  '',
+].join('\n');
+const maxBodyChars = MAX_CHARS - FOOTER.length;
+
+if (maxBodyChars <= 0) {
+  throw new Error('MAX_CHARS is too small to fit the required footer.');
+}
+
+const content = await fs.readFile(LLMS_FILE, 'utf8');
```
banafshr: this is also missing error handling. the risk is that the CI would fail without any messages. it's minimal. up to you if you want to add it

scripts/limit-llms-txt-size.mjs
```diff
@@ -0,0 +1,41 @@
...
+const lines = content.split('\n');
+
+const includedLines = [];
+let charCount = 0;
+
+for (const line of lines) {
```
banafshr: I'm wondering if the truncating here might cause an issue with section headers? for example a section header might get truncated 🤔

harshini-rangaswamy: Good point. The script already truncates by full lines, so it should not cut a heading in the middle. I added a small cleanup step to handle the other case where a heading is left at the end without any content under it. Tested this locally and works.

banafshr (approved):

PR #1524, opened 2026-05-11
update: add mcp tool grouping option + enable read-only by default

tomershay:
1. Add documentation for the new tool scoping feature, allowing users to choose whether to install all MCP tools or only specific tool groups such as PG, Kafka, etc. The default behavior remains installing all tools.

2. Enable read-only mode by default, encouraging users to start with safer read-only access.

3. Remove the limited availability info notification, since the same information already appears next to the title. Keeping both is redundant and takes up too much above-the-fold space.


harshini-rangaswamy (approved):
LGTM! Thanks for the update.

yoni-nezer-aiven (approved):
lgtm

PR #1543, opened 2026-05-13
update: Fix call-out about Kafka OAUTHBEARER and SASL config

aiven-anton:
The note claimed that OAUTHBEARER requires at least one of PLAIN, SCRAM-SHA-256, or SCRAM-SHA-512 to also be enabled. That's wrong — kafka.sasl_oauthbearer_jwks_endpoint_url alone is sufficient and the other SASL mechanisms can be disabled independently.

The requirement to enable kafka_authentication_methods.sasl is accurate and kept.

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/products/kafka/howto/enable-oidc.md
+++ b/docs/products/kafka/howto/enable-oidc.md
@@ -18,8 +18,8 @@ of the Kafka brokers. This restart does not cause service downtime.
 
 :::note
 To use the `OAUTHBEARER` mechanism, you must enable `kafka_authentication_methods.sasl`.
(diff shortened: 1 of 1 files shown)


HelenMel (commented):
Are you sure? Are you sure a sasl shouldn't be enabled first?(you removed that line)

aiven-anton:
@HelenMel Ah, not sure, true. Will need to defer this.

HelenMel (approved):

harshini-rangaswamy (approved):
Thanks for this change!

PR #1544, opened 2026-05-13
fix: broken upstream Apache Kafka documentation links

aiven-anton:
Some links were referencing very old and obsolete versions -- those are updated to the most recent upstream stable version. Others are entirely broken due to recent restructuring in the upstream documentation, those are addressed and now lead to their new correct URLs.

## Checklist

- [ ] I checked the output.


harshini-rangaswamy (approved):
Thanks for fixing these :)

PR #1568, opened 2026-05-21
feat: PG Studio execute multiple queries

wojcik-dorota:
https://aiven.atlassian.net/browse/PGDX-91


docs/products/postgresql/howto/pg-studio/write-run-queries.md
```diff
@@ -43,16 +45,42 @@ This is useful for:
...
+   SELECT * FROM users LIMIT 10;
+   ```
+
+1. Click **Run**.
+1. View the results in the results panel. When more than one statement runs, each result
+   appears in its own tab labeled **Query 1**, **Query 2**, and so on:
+   - A red error indicator on a tab means the statement encountered an error. If a
+     statement fails, execution stops and remaining statements are not run. Results from
+     statements that completed before the error are preserved.
+   - When only one statement runs, the tab bar is hidden automatically.
```
nicosierra: I am not acquainted with our docs, but this level of information I'd say is not really relevant.

docs/products/postgresql/howto/pg-studio/write-run-queries.md
```diff
@@ -43,16 +45,42 @@ This is useful for:
 - Testing each subquery independently before combining a complex query
 - Verifying data in a specific table without executing setup or cleanup statements
 
+## Run multiple queries
+
+You can write several SQL statements in the editor and run them all at once. Separate each
+statement with a semicolon. When you click **Run**, PG Studio executes all statements in
```
nicosierra: I'd skip the `Separate each statement with a semicolon.` as there are other ways of having multiple statements.

docs/products/postgresql/howto/pg-studio/write-run-queries.md
```diff
@@ -43,16 +45,42 @@ This is useful for:
...
+the current tab sequentially and displays the results in separate tabs.
+
+1. In the SQL editor, write your SQL statements, separating each with a semicolon:
```
nicosierra: same here, i'd skip the text after the comma.

docs/products/postgresql/howto/pg-studio/write-run-queries.md
nicosierra: namely,
```
   - When only one statement runs, the tab bar is hidden automatically.
```

nicosierra:
One note, the "Run" button will only run all statements if all are selected, otherwise it defaults to the one where the text cursor sits in

wojcik-dorota:
> One note, the "Run" button will only run all statements if all are selected, otherwise it defaults to the one where the text cursor sits in

Updated :-)

nicosierra (approved):
🚀

adir-pur (approved):

PR #1573, opened 2026-05-25
feat: integrate posthog with docs

tomershay:
Integrate Posthog with Docs:

  - Integrate PostHog event tracking and session replay into the docs site, gated by OneTrust cookie consent.
  - Tracking is disabled by default and only enabled if the user accepts cookies via OneTrust group 115 (UX/Functional cookies).
  - Add a "Cookie settings" link in the footer so users can revisit their consent choices (matches the marketing site).


banafshr (approved):
LGTM!

harshini-rangaswamy (approved):

roman-pozd (approved):

PR #1580, opened 2026-05-27
fix: load kapa ai as async to not block render

banafshr:
## Describe your changes

Before
Screenshot 2026-05-27 at 10 19 33

after (this is slower because it's a preview environment and not production)
Screenshot 2026-05-27 at 10 22 38


harshini-rangaswamy (approved):
Thanks for this update!

PR #1590, opened 2026-05-29
update: PG controlled switchover

wojcik-dorota:
https://aiven.atlassian.net/browse/OPD-1109

CHANGES

- COMPLETED_EARLY state added to the state-values list — switchover that finishes outside a controlled window (for example, when the old primary fails during the wait).
- completed_at field — time the switchover completed; can be null.
- termination_reason field — reason for completion, such as Controlled switchover or old generation node timed out; can be null.
- Early-completion clarification — a sentence explaining that if old nodes time out or fail outside a configured window, Aiven promotes and completes early without waiting, setting state to COMPLETED_EARLY.
- Added a short field-reference list describing the whole controlled_switchover object (it previously only explained scheduled_start_time), and updated the JSON example to a COMPLETED response showing the two new fields populated.


giacomo-alzetta-aiven (approved):
LGTM!

PR #1606, opened 2026-06-01
fix: default to claude code in mcp and add code block copy button

tomershay:
After reviewing last 50 user session recordings:

- Made Claude Code the first/default tab in the MCP install instructions, as most users seem to be using Claude Code.
- Added a copy to clipboard button on each code block.

### DIFF
--- a/docs/tools/mcp-server.md
+++ b/docs/tools/mcp-server.md
@@ -30,27 +30,7 @@ can sign in and authorize MCP access.
 ## Install the Aiven MCP {#configure-aiven-mcp}
 
 <Tabs groupId="aiven-mcp-clients">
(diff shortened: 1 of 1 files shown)


yonatan-dvir (approved):
LGTM 
Just wondering - the copy button was already there, isn't it?

tomershay:
@yonatan-dvir thanks for the review. Yes, initially I changed the copy to clipboard to show up constantly, not only on-hover, as I think many users missed that it exists (and so did I initially).

But instead of hacking it specifically for this page, I'll raise it to the technical writers team to see if we can/should change the behavior for all code editors in the docs website. I think we can merge this PR without it, and I'll follow up with it. thanks

harshini-rangaswamy (approved):

PR #1613, opened 2026-06-02
fix: add retry mechanism to fix posthog loading race condition

tomershay:
## Summary
  - Fixes `window.posthog.startSessionRecording is not a function` in Firefox.
  - Cause: our consent code sometimes ran before PostHog's library finished loading, so `startSessionRecording` didn't exist yet on the stub.
  - Fix: retry until PostHog is fully loaded, then apply consent.

### DIFF
--- a/static/page_scripts/tracking.ts
+++ b/static/page_scripts/tracking.ts
@@ -13,15 +13,19 @@ declare global {
   }
 }
 
-function applyPosthogConsent() {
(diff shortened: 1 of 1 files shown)


yoni-nezer-aiven (approved):

staceysalamon-aiven (approved):
Thank you!

PR #1619, opened 2026-06-03
fix: set correct path for font

banafshr:
## Describe your changes

`baseURL` changes depending on the environment. the current font path did not take that into account. 
moved css styles into the head directly to set the correct baseurl for the path

https://aiven.atlassian.net/browse/BRCOM-1264

## Checklist

- [ ] I checked the output.


staceysalamon-aiven (approved):
Thanks!

PR #1648, opened 2026-06-12
update: MySQL doc IA reorg

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-1912

[DOC PREVIEW](https://doc-1912-replace-diataxis-wi.aiven-docs.pages.dev/docs/products/mysql)


UX-Irene (approved):
Looks good. Only thing I think we need to tackle at some point is the Service Management and Cluster Section. But until we have settled on the standard structure for those sections, lets leave this as is.

PR #1651, opened 2026-06-14
update: Grafana docs IA reorg

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-1915


UX-Irene (dismissed):
I noticed that there are two Login section, on standalone and repeated under the Getting started. Please consider merging this into the Getting started section. Seems like the Standalone one is talking about OpenSearch so is a bit confusing. Perhaps enough to have just under getting started.

UX-Irene (approved):
approve

PR #1659, opened 2026-06-15
update: add ClickHouse 25.8 default settings reference

harshini-rangaswamy:
## Describe your changes

Preview: https://fleech-220-document-settings.aiven-docs.pages.dev/docs/products/clickhouse/reference/25-8-default-settings

## Checklist

- [x] I checked the output.

### DIFF
--- a/docs/products/clickhouse/reference/25-8-default-settings.md
+++ b/docs/products/clickhouse/reference/25-8-default-settings.md
@@ -16,8 +16,12 @@ Each setting includes an Aiven default value and indicates whether you can chang
 
 - **Can be changed: No**: Aiven manages the setting value. You cannot override it.
 - **Can be changed: Yes**: Aiven sets the default value, but you can change it
-  where the setting is configurable. Where a setting has limits, such as an
-  allowed range or minimum value, the **Description** column shows them.
+  where the setting is configurable.
+
+Where Aiven defaults vary by service plan or node resources, the **Aiven default**
+column shows ranges or placeholders. The **Description** column explains why the
+default differs or lists any configurable bounds, such as an allowed range or
+minimum value.
 
 Values shown as ranges or placeholders, such as
 `[3..80, depending on CPU count]`, `{cpu_count}`, or `[~7% of RAM]`, are sized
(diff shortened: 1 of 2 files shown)


smirnoffs (approved):
With the release of 26.3, we will have to think about how to mention version-specific differences, and we will have to change the title. For now, it's good to go.

docs/products/clickhouse/reference/25-8-default-settings.md
```diff
@@ -0,0 +1,150 @@
...
+| `readonly`                                              | `0`                                     | Yes            | Controls read/write access for the session. The default allows writes.                                                                                                          |
+| `stream_like_engine_allow_direct_select`                | `1`                                     | Yes            | Allows direct `SELECT` queries on stream-like engines, which can have suboptimal performance or behavior. A future release returns this default to the upstream default.         |
+| `write_full_path_in_iceberg_metadata`                   | `1`                                     | Yes            | Writes full Iceberg metadata paths for compatibility with mainstream ClickHouse behavior.                                                                                        |
+
+<!-- markdownlint-enable MD013 MD060 -->
+
+:::note
+The `readonly` setting applies to your session only. Separately, Aiven can switch
```
Khatskevich: I am not sure if it makes sense to not that the `readonly` is applied to your session only, because all the settings work the same (apply to your session only). But the key element here is that Aiven can turn "readonly=0 Can be changed=True" to "readonly=1 Can be changed=False"

docs/products/clickhouse/reference/25-8-default-settings.md
```diff
@@ -0,0 +1,150 @@
...
+defaults in version 25.8, grouped by session, table, and server scope.
+
+## How to read the settings tables
+
+Each setting includes an Aiven default value and indicates whether you can change it:
+
+- **Can be changed: No**: Aiven manages the setting value. You cannot override it.
+- **Can be changed: Yes**: Aiven sets the default value, but you can change it
+  where the setting is configurable. Where a setting has limits, such as an
+  allowed range or minimum value, the **Description** column shows them.
```
Khatskevich: I guess limits are shown in the "Aiven default" column. And the "description" just clarifies why did we change the default value

docs/products/clickhouse/reference/25-8-default-settings.md
```diff
@@ -0,0 +1,150 @@
...
+| `threadpool_writer_pool_size`                   | `[3..80, depending on CPU count]`                   | Scales filesystem writer concurrency with service CPU capacity.                                                                                          |
+| `uncompressed_cache_size`                       | `[~7% of RAM]`                                      | Sizes the uncompressed block cache based on the managed memory budget.                                                                                   |
+| `user_with_indirect_database_creation`          | `avnadmin`                                          | Restricts indirect database creation privileges to the managed admin user.                                                                               |
+| `vector_similarity_index_cache_size`            | `[10% of RAM]`                                      | Sizes the vector similarity index cache based on the managed memory budget.                                                                              |
```
Khatskevich: 1. Oh, actually this setting is the only one (so far) which can be directly overwritten by client.
It is 7% by default only
2. Sorry, there is a mistake. We should write "7% of RAM" (if was 10% of service-usable ram, which confused me)

Image

Khatskevich (changes requested):

docs/products/clickhouse/reference/25-8-default-settings.md
smirnoffs: The default used to be 0.1, wasn't it?

Khatskevich (approved):
looks good

PR #1694, opened 2026-06-24
update: fix claude desktop connection info

roman-pozd:
## What

Fixes the Claude Desktop connection setup for the Aiven MCP docs.

- **Hosted Claude Desktop tab**: use the `mcp-remote` bridge instead of the native `type: "http"` connector, so the read-only, allow_secrets, scopes, and marketplace toggles attach their URL params dynamically — matching the other client tabs.
- **Local installation (stdio)**: document `AIVEN_ALLOW_SECRETS` across the Claude Code, Cursor, Claude Desktop, and VS Code tabs.

## Why

The native HTTP connector didn't expose a working re-auth path in some Claude Desktop builds and didn't carry the configurable URL params. The `mcp-remote` bridge resolves both. `AIVEN_ALLOW_SECRETS` was added to the stdio MCP and was previously undocumented.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/tools/mcp-server.md
+++ b/docs/tools/mcp-server.md
@@ -73,7 +73,7 @@ For more information, see the [Cursor MCP documentation](https://cursor.com/docs
    <MCPConfigSection
      baseUrl={mcpUrl}
      format="json"
-     configTemplate={(url) => ({mcpServers: {aiven: {type: "http", url}}})}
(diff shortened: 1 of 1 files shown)


harshini-rangaswamy (approved):
Thanks for this update!

PR #1698, opened 2026-06-25
update: set read_only = false as a default

roman-pozd:
updating the default of read_only flag to flase for the mcp installation


harshini-rangaswamy (approved):
Thanks for the update!

PR #1710, opened 2026-06-29
update: service management reuse for DB services [part 1]

wojcik-dorota:
Add per-service Service management articles from reusable includes ([DOC-1947, part 1](https://aiven.atlassian.net/browse/DOC-1947))

1. Reuse

Adds per-service "Service management" articles for PostgreSQL, MySQL, OpenSearch, Valkey, Grafana, and Metrics, built from the shared components in static/includes (tags, power on/off + delete, rename, fork, backups, service users). Each article adds an intro sentence, a Related pages section, and service-specific details on top of the reused content.

Articles are added only where the service supports the feature, verified against aiven-core schemas:
- Fork + Rename: all except Metrics (Thanos isn't forkable)
- PITR fork: PostgreSQL, MySQL only
- Backups (schedule): PostgreSQL, MySQL, Valkey
- Service users: PostgreSQL, MySQL, OpenSearch, Valkey

2. Service management subcategorization

- Service operations → Power on/off, Rename, Use resource tags, Fork, (Service users where it exists), and Manage parameters → Advanced parameters
- Upgrades and maintenance → the service's version-upgrade page (PG upgrade, MySQL manage-mysql-version, OpenSearch os-version-upgrade, Valkey valkey-version-upgrade; Grafana has none, so skipped)
- Capacity and scaling → existing memory/tuning pages (PG pg-shared-buffers; MySQL memory + tuning; Valkey memory + overcommit; OpenSearch/Grafana have none, so skipped)
- Direct items kept as-is → PG: create-database, enable-jit, pg_repack, repair-pg-index, wraparound; OpenSearch: dedicated-node-roles, HA, plugins, plugin-list, limitations; Grafana: send-emails, plugins; Valkey: restricted-commands

3. Misc
- Replaces grafana/howto/pitr-process-for-grafana with the new Grafana fork article and adds a 301 redirect.
- Removes "Cluster management" sections as redundant and overlapping with service management reusables.

[DOC PREVIEW](https://doc-1947-db-part-1-service-m.aiven-docs.pages.dev/docs/products/postgresql)


UX-Irene:
Overall looking more consistent. I can see that some of the generic pages have not yet been added to the service pages, so lets revisit once we have all components ready. Made new notes in the Lucid board. 
One thing:
For PG, I think it would be good to move the page "DB, table, index size" from Observability section to the new Manage Databases section. Dont you think it makes more sense there, or am I wrong?

wojcik-dorota:
> Overall looking more consistent. I can see that some of the generic pages have not yet been added to the service pages, so lets revisit once we have all components ready. Made new notes in the Lucid board. One thing: For PG, I think it would be good to move the page "DB, table, index size" from Observability section to the new Manage Databases section. Dont you think it makes more sense there, or am I wrong?

Thanks, @UX-Irene ! Good catch 👍🏻
How about moving it to "Scaling and performance", which currently holds: pg-disk-usage, prevent-full-disk, and pg-shared-buffers?
Reasoning: "DB, table, index size" is about disk space usage of database objects, the same theme as pg-disk-usage and prevent-full-disk. It even cross-links to pg-disk-usage and notes WAL contribution to disk usage. Checking object sizes is usually done prior to diagnosing disk pressure and planning capacity.

UX-Irene:
Moving it to "Scaling and performance" sound like the right decision.

UX-Irene (approved):

PR #1736, opened 2026-07-03
update: byoc oci and azure restrictions

wojcik-dorota:
https://aiven.atlassian.net/browse/IP-1080


aiven-timoe (approved):
Perfect - great clarification

PR #1739, opened 2026-07-05
fix: add mcp ai clients icons and collapse advanced settings

tomershay:
Experience improvements for the MCP docs page:

* Collapse the advanced settings so it will be easier for users to install the MCP with default settings
* Enable read only mode by default
* Add icons for the AI clients to make the tabs look more appealing

### DIFF
--- a/docs/tools/mcp-server.md
+++ b/docs/tools/mcp-server.md
@@ -9,6 +9,14 @@ import TabItem from '@theme/TabItem';
 import {mcpUrl} from '@site/src/components/mcpAivenLiveConstants';
 import MCPConfigSection from "@site/src/components/MCPConfigSection";
 import CursorConfigTab from "@site/src/components/CursorConfigTab";
+import McpClientTabLabel from "@site/src/components/McpClientTabLabel";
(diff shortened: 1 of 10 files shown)


src/components/CursorConfigTab/index.tsx
```diff
@@ -9,7 +9,7 @@ type CursorConfigTabProps = {
 };
 
 export default function CursorConfigTab({ baseUrl }: CursorConfigTabProps): JSX.Element {
-  const [state, setState] = useState<MCPConfigState>({ readOnly: false, scopes: [], marketplace: '', allowSecrets: false });
+  const [state, setState] = useState<MCPConfigState>({ readOnly: true, scopes: [], marketplace: '', allowSecrets: false });
```
roman-pozd: We've changed this to false as we've got a lot of feedbacks from the hackathon that people just copy pasted the url to install the mcp and stuck for hours figuring out thats it's a limited functionality and they actually can enable the write functionality via a param. 🤔

roman-pozd (approved):

harshini-rangaswamy (approved):
Thanks for the improvements!

PR #1740, opened 2026-07-05
update: add write_allowlist param to mcp

roman-pozd:
we're adding write_allowlist that can be used beside read_only=true and provide the ability to exlucde specific write tools in read only mode.


harshini-rangaswamy (approved):
Thanks for this!

PR #1769, opened 2026-07-15
fix: Broaden tiered storage plan limitation to all startup plans

Stuzanna:
Updates tiered storage documentation to clarify that the feature is unavailable on all startup plans, not just startup-2 plans.


harshini-rangaswamy (approved):
Thanks for this!

PR #1800, opened 2026-07-22
update: EOL/EOA date update automation

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-1976

Run the script locally using `python3 scripts/update-eol-dates.py`.


scripts/update-eol-dates.py
```diff
@@ -0,0 +1,377 @@
...
+    for i in range(num_cols):
+        cells.setdefault(i, PLACEHOLDER)
+    return "| " + " | ".join(cells[i] for i in range(num_cols)) + " |\n"
+
+
+def process_tools_table(
+    table_rows: list[str], config: ToolsConfig, github_majors: set[str],
+    today: str, num_cols: int,
+) -> tuple[list[str], bool]:
+    """Detect new major versions; set preceding version's EOL to today if it was a placeholder."""
```
staceysalamon-aiven: There's no way to get the date from the repos for these, and it wouldn't be a good idea to set it to the day this runs since that's unlikely to be accurate. I think for the dev tools, this probably can't be easily automated.

wojcik-dorota: Thanks! I removed the tools part. Now it only updates the service lifecycle dates.

.github/workflows/update-eol-dates.yaml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          python-version: '3.11'
+      - name: Run EOL date update script
+        # Exit code 0 = no changes; 1 = doc updated; 2 = error
+        run: |
+          python scripts/update-eol-dates.py
+          echo "SCRIPT_EXIT=$?" >> "$GITHUB_ENV"
+        continue-on-error: true
+      - name: Create pull request
+        if: env.SCRIPT_EXIT == '1'
```
harshini-rangaswamy, line 30: `scripts/update-eol-dates.py` exits with `0` for no changes, `1` when the document is updated, or `2` on error, as documented in its docstring and `main()`. However, this workflow step doesn't capture those exit codes correctly. GitHub Actions runs `run:` steps with `bash -e` by default, so the step stops after the `python` command when the script exits with `1` or `2`. The `echo` command never runs, `SCRIPT_EXIT` remains unset, and neither the "Create pull request" nor the "Fail if script errored" step is triggered.

Could we capture the exit code explicitly before `-e` can kill the step? 
May be worth checking this with Bani. 

```yaml
run: |
  set +e
  python scripts/update-eol-dates.py
  script_exit=$?
  echo "SCRIPT_EXIT=$script_exit" >> "$GITHUB_ENV"
```

wojcik-dorota: Thanks for catching this, @harshini-rangaswamy 💪🏻 Fixing [here](https://github.com/aiven/aiven-docs/pull/1800/changes/edd449efa0f026e995a4ac2104e20ea00caa3667) 🤗

@banafshr , could you please also have a look 🙏🏻

harshini-rangaswamy (approved):
Looks good, thanks!

.github/workflows/update-eol-dates.yaml
```diff
@@ -0,0 +1,43 @@
+name: Update EOL dates
+
+on:
+  schedule:
+    - cron: '0 7 * * 1'  # weekly on Monday at 7 AM UTC
+  workflow_dispatch:
+
+jobs:
+  update-eol-dates:
+    name: Update EOL dates from Aiven API
```
banafshr, line 10: how fast/slow is the aiven api? i'm thinking if there's a possibility of this failing if the api is very slow. maybe we can add a timeout here?

.github/workflows/update-eol-dates.yaml
```diff
@@ -0,0 +1,43 @@
...
+          body: |
+            > [!NOTE]
+            > Auto-generated PR. Dates synced from `api.aiven.io/v1/service_versions`.
+            > Review the diff and merge when accurate.
+          base: main
+          branch: eol-dates-update
+          labels: automated-pr, review/TW
+          delete-branch: true
+      - name: Fail if script errored
+        if: env.SCRIPT_EXIT == '2'
```
banafshr, line 46: would it make sense to have one for SCRIPT_EXIT = 0? just to log that there was no changes

banafshr: also, what happens if the script returns an unexpected error? maybe would be good to have a generic workflow too, to log if there's an issue

banafshr (approved):

PR #1824, opened 2026-07-29
update: Valkey backups taken with persistence off and backup_hour set

wojcik-dorota:
https://aiven.atlassian.net/browse/VK-270


proddata:
I am afraid we might need to revisit this, as it might make more sense to change the actual behaviour, instead of changing the documentation.

proddata (approved):

PR #1828, opened 2026-07-30
update: Valkey defragmentation and memory management

wojcik-dorota:
https://aiven.atlassian.net/browse/VK-273

Added a new "Memory fragmentation and active defragmentation" section explaining why fragmentation grows on long-running, high-churn services, documenting the new valkey_activedefrag config option (off by default, main-thread CPU cost), and recommending testing on a forked service first.

### DIFF
--- a/docs/products/valkey/concepts/memory-usage.md
+++ b/docs/products/valkey/concepts/memory-usage.md
@@ -151,9 +151,7 @@ system.
 
 :::note
 Active memory defragmentation is disabled by default. It runs on the main thread and
-consumes CPU, so it can increase latency under load. Test the option on a
-[forked service](/docs/products/valkey/howto/fork-service) before enabling it on a
-production service.
+consumes CPU, so it can increase latency under load.
 :::
 
 For the full list of configuration options, including `valkey_activedefrag`, see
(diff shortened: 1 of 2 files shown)


docs/products/valkey/concepts/memory-usage.md
```diff
@@ -137,6 +137,28 @@ If you frequently need to write large volumes of data, contact Aiven support to
...
+To reduce fragmentation, enable the **Active memory defragmentation**
+(`valkey_activedefrag`) advanced configuration option. When enabled, Valkey relocates
+objects off sparsely used memory pages and returns the freed memory to the operating
+system.
+
+:::note
+Active memory defragmentation is disabled by default. It runs on the main thread and
+consumes CPU, so it can increase latency under load. Test the option on a
+[forked service](/docs/products/valkey/howto/fork-service) before enabling it on a
+production service.
```
bogdanp05: I'd probably remove `Test the option on a forked service...`. The problem is that fragmentation happens after a series of writes and deletes. If you have a service with a high fragmentation ratio and you fork it to a new service:
1. The new service will have a low fragmentation ratio because when it loads the data into memory for the first time, jemalloc will just allocate contiguous memory blocks.
2. Even if you artificially create a high fragmentation ratio by running random key deletes, you'd also need to reproduce the actual prod workload to notice any difference in throughput.

bogdanp05 (commented):
Looks good, minor comment.
I'll let you know once the feature is deployed so we can merge this.

bogdanp05 (approved):
The feature is deployed, this can be merged

PR #1857, opened 2026-08-10
update: service version lifecycle and EOL/EOA reusable components

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-1996


harshini-rangaswamy (approved):
Looks good, thanks!

PR #1866, opened 2026-08-13
update: upgrade Docusaurus to 3.10.2

harshini-rangaswamy:
## Describe your changes

- Upgrade official `@docusaurus/*` packages from 3.7.0 to 3.10.2 (last release in the 3.x line). See the [3.10 release notes](https://docusaurus.io/blog/releases/3.10).
- Enable Docusaurus Faster with the stable `future.faster` flag (replaces `experimental_faster`). Faster now requires `future.v4.removeLegacyPostBuildHeadAttribute`, so that one v4 flag is turned on as well.
- We did **not** set `v4: true`. That enables all v4 prep flags, including stricter MDX, which breaks existing heading IDs (`## Title {#custom-id}`) and HTML comments (`<!-- vale off -->`). Migrating those is a follow-up, not part of this upgrade.
- Also bumped Yarn (4.7.0 → 4.18.0) and the packages included in `yarn run aiven-upgrade` (Algolia client, Aquarium). `@docusaurus/theme-classic` was bumped separately because the upgrade script does not include it.

## Checklist

- [ ] I checked the output.

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -28,7 +28,12 @@ const posthogPlugin: PluginConfig = process.env.POSTHOG_PUBLIC_KEY
 
 const config: Config = {
   // Testing faster build
-  future: {},
+  future: {
+    v4: {
+      removeLegacyPostBuildHeadAttribute: true,
+    },
+    faster: true,
+  },
   title: 'Aiven docs',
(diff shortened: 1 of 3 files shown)


wojcik-dorota:
Thank you, @harshini-rangaswamy 🙏🏻

PR #1886, opened 2026-08-19
feat: secret redaction in API

staceysalamon-aiven:
## Checklist

- [ ] I checked the output.

### DIFF
--- a/docs/tools/api/secret-redaction.md
+++ b/docs/tools/api/secret-redaction.md
@@ -5,16 +5,16 @@ sidebar_label: Secret redaction
 
 Service user passwords, service `user_config` fields, and integration endpoint secrets are redacted in API responses by default.
 
-If your integrations require access to secrets, ensure your API calls include
-the `include_secrets=true` query parameter and that the caller has a
+If your integrations require access to secrets, include the `include_secrets=true`
+query parameter in the API calls. The caller also needs to have a
 [role or permission](/docs/platform/concepts/permissions) that allows them to read secrets.
 
 The following is a summary of the permissions required for each secret type:
 
 - **Service user passwords**: `admin`, `operator`, `developer`,
    `service:secrets:read`, or `service:users:write`
-- **Service user_config secrets**: `admin`, `operator`, or `service:secrets:read`
-- **Integration-endpoint secrets**: `admin`, or `project:integrations:write`
+- **Service `user_config` secrets**: `admin`, `operator`, or `service:secrets:read`
+- **Integration endpoint secrets**: `admin` or `project:integrations:write`
 
 Only the `GET` endpoints listed can reveal secrets.
 Write endpoints and the integration endpoint list always redact secrets.
--- a/docs/tools/api/secret-redaction.md
+++ b/docs/tools/api/secret-redaction.md
@@ -1,0 +1,62 @@
+---
+title: Secret redaction in Aiven API
+sidebar_label: Secret redaction
+---
+
+Service user passwords, service `user_config` fields, and integration endpoint secrets are redacted in API responses by default.
+
+If your integrations require access to secrets, ensure your API calls include
+the `include_secrets=true` query parameter and that the caller has a
+[role or permission](/docs/platform/concepts/permissions) that allows them to read secrets.
+
+The following is a summary of the permissions required for each secret type:
+
+- **Service user passwords**: `admin`, `operator`, `developer`,
+   `service:secrets:read`, or `service:users:write`
+- **Service user_config secrets**: `admin`, `operator`, or `service:secrets:read`
+- **Integration-endpoint secrets**: `admin`, or `project:integrations:write`
+
+Only the `GET` endpoints listed can reveal secrets.
+Write endpoints and the integration endpoint list always redact secrets.
+
+The following endpoints include secrets:
+
+- **Service user passwords**
+
+  - `GET /project/{project}/service/{service_name}/user/{service_username}`
+
+  - `GET /project/{project}/service/{service_name}`
+
+  - `GET /project/{project}/service`
+
+  - `POST /project/{project}/service`
+
+  - `PUT /project/{project}/service/{service_name}`
+
+  - `PUT /project/{project}/service/{service_name}/user/{service_username}/credentials/reset`
+
+  - `PUT /project/{project}/service/{service_name}/user/{service_username}`
+
+  - `PATCH /project/{project}/service/{service_name}/service_type`
+
+- **Service user configuration**
+
+  - `GET /project/{project}/service/{service_name}`
+
+  - `GET /project/{project}/service`
+
+  - `POST /project/{project}/service`
+
+  - `PUT /project/{project}/service/{service_name}`
+
+  - `PATCH /project/{project}/service/{service_name}/service_type`
+
+- **Integration endpoint secrets**
+
+  - `GET /project/{project}/integration_endpoint/{integration_endpoint_id}`
+
+  - `GET /project/{project}/integration_endpoint`
+
+  - `POST /project/{project}/integration_endpoint`
+
+  - `PUT /project/{project}/integration_endpoint/{integration_endpoint_id}`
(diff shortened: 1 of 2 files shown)


docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Secret redaction in Aiven API
+sidebar_label: Secret redaction
+---
+
+Service user passwords, service `user_config` fields, and integration endpoint secrets are redacted in API responses by default.
```
rsalevsky: Only **secret** `user_config` fields are redacted, not every `user_config` field. Also say that redacted values show as `<redacted>` — integrations that round-trip config need that placeholder (or they overwrite the secret).

docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Secret redaction in Aiven API
+sidebar_label: Secret redaction
+---
+
+Service user passwords, service `user_config` fields, and integration endpoint secrets are redacted in API responses by default.
+
+If your integrations require access to secrets, include the `include_secrets=true`
+query parameter in the API calls. The caller also needs to have a
+[role or permission](/docs/platform/concepts/permissions) that allows them to read secrets.
```
rsalevsky: This sentence is doing too much, and it implies `include_secrets=true` works on every call. In acorn it is a **GET** query parameter only; create/update never read it.

Also document the 403: granular callers without a secrets permission get `403` and `"User does not have permission to read service secrets"`. Legacy `read_only` stays `200` with redacted values.

Suggested:

```markdown
To read secrets, send `include_secrets=true` on a GET request and use a
[role or permission](/docs/platform/concepts/permissions) that can read
that secret type. Callers without that permission receive `403 Forbidden`.
```

docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,62 @@
...
+If your integrations require access to secrets, include the `include_secrets=true`
+query parameter in the API calls. The caller also needs to have a
+[role or permission](/docs/platform/concepts/permissions) that allows them to read secrets.
+
+The following is a summary of the permissions required for each secret type:
+
+- **Service user passwords**: `admin`, `operator`, `developer`,
+   `service:secrets:read`, or `service:users:write`
+- **Service `user_config` secrets**: `admin`, `operator`, or `service:secrets:read`
+- **Integration endpoint secrets**: `admin` or `project:integrations:write`
```
rsalevsky: Granular set is right; a few caveats worth spelling out:

- `developer` alone cannot read `user_config` secrets (needs `service:secrets:read`).
- `operator` cannot unredact **integration-endpoint** secrets — only `admin` or `project:integrations:write`.
- Two-item list: `admin` or `project:integrations:write` (drop the comma after `admin`).
- Align the label with the heading below (`user_config` vs “user configuration”; drop the hyphen in “Integration-endpoint”).

Org-level roles (`role:organization:admin`, `role:project:admin`, `role:project:manage`) also grant some of these. If public docs should only list Console-facing project roles, say so and omit the `role:*` names.

docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,62 @@
...
+
+Only the `GET` endpoints listed can reveal secrets.
+Write endpoints and the integration endpoint list always redact secrets.
+
+The following endpoints include secrets:
```
rsalevsky: These two sentences contradict the list that follows. “The following endpoints include secrets” mixes the *affected* surface with the *reveal* surface.

`include_secrets` is GET-only. Write endpoints call `get_service_for_user()` / `_should_return_secrets()` with `include_secrets=None`, so they cannot opt in. The integration-endpoint **list** also never reads the query param.

Split into two lists:

- **Can return plaintext** (GET + `include_secrets=true` + permission): service GET, service list, service-user GET, integration-endpoint GET.
- **Always redacted:** service create/update, credentials reset, user update, service-type PATCH, integration-endpoint create/update, and the integration-endpoint list.

Exception: `POST /v1/project/{project}/service/{service_name}/user` still returns the new password in plaintext and is not listed. Include it as an exception, or drop the absolute “always”.

Suggested lead-in:

```markdown
`include_secrets` is a GET query parameter. Write endpoints and the
integration endpoint list always return redacted values.
```

docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,62 @@
...
+
+  - `PUT /project/{project}/service/{service_name}`
+
+  - `PATCH /project/{project}/service/{service_name}/service_type`
+
+- **Integration endpoint secrets**
+
+  - `GET /project/{project}/integration_endpoint/{integration_endpoint_id}`
+
+  - `GET /project/{project}/integration_endpoint`
```
rsalevsky: This list endpoint never reads `include_secrets`. With redaction enabled it always returns `<redacted>`. Keep it in the “always redact” list, not under endpoints that can include secrets.

docs/tools/api/secret-redaction.md
rsalevsky, line 1: File-level notes:

- This file is committed with CRLF (`\r\n`). Convert to LF.
- Paths omit `/v1`. The rest of the API docs and `api.aiven.io` use `/v1/project/...`.
- Add headings, a curl example, `RelatedPages`, and a link from the API landing page. Suggested shape: (1) what is redacted, (2) how to request secrets, (3) permissions, (4) reveal vs always-redact endpoints, (5) round-tripping `<redacted>`.
- Fill in the empty PR description and checklist.

rsalevsky (changes requested):
Requesting changes — the page is a useful start (and the sidebar category is fine), but it currently contradicts itself on which endpoints can return secrets.

Checked against aiven-core: `aiven/acorn/service_secrets.py`, `aiven/acorn/service_apis.py`, `aiven/acorn/service_integrations/apis.py`, and `tests/local/acorn/test_secret_redaction.py`.

Main fixes:
- Split “affected” vs “can reveal” endpoints (`include_secrets` is GET-only).
- Don’t list the integration-endpoint list as able to return secrets.
- Call out service-user create (still returns the password in plaintext).
- Document `<redacted>`, the 403, `/v1` paths, and the operator/developer permission caveats.
- Convert the new file from CRLF to LF.

docs/tools/api/secret-redaction.md
staceysalamon-aiven: I've decided to omit this because we have separate documentation for the permissions and this makes it easier to maintain if/when things change with permissions.

docs/tools/api/secret-redaction.md
staceysalamon-aiven: I'm not convinced this is contradictory, but I've removed the list of endpoints that always redact. As with permissions, I think it'll be too difficult to maintain as write endpoints are added/changed.

docs/tools/api/secret-redaction.md
rsalevsky: If we mention the old admin, operator and developer then for completness we should mention the org level roles as well. Especially as e.g. role:project:admin is a direct replacement for admin.

rsalevsky (changes requested):

docs/tools/api/secret-redaction.md
```diff
@@ -0,0 +1,33 @@
...
+leaving this placeholder prevents accidental overwriting of the secret.
+
+To read secrets on a `GET` request:
+
+- Send `include_secrets=true` as a query parameter.
+- Use a [role or permission](/docs/platform/concepts/permissions) that can read
+  that secret type.
+
+For service secrets, calls without the required permissions receive a `403 Forbidden`
+response. For integration endpoint secrets, calls return redacted values.
```
staceysalamon-aiven, line 18: @rsalevsky I added this sentence about integration endpoints since it looks like it doesn't return 403 for these, but please correct me if I'm wrong about this. 🙏

rsalevsky: That's right. Integration-endpoint GET never 403s for secrets.

`_should_return_secrets()` returns `false` when the caller lacks `service_integration_endpoint_secrets_read`, and the handler redacts in place. Tests for `operator` / `project:integrations:read` with `include_secrets=true` still get `200` and `<redacted>`.

Service GET/list is the one that 403s (granular callers with no secrets permission). Legacy `read_only` is the exception there: `200` with redacted values, not 403.

rsalevsky (approved):

PR #1887, opened 2026-08-19
feat: project read access role

staceysalamon-aiven:
## Checklist

- [ ] I checked the output.


docs/platform/concepts/permissions.md
```diff
@@ -79,14 +79,15 @@ These permissions apply to the
...
...rvice tags for all services in the project. </li> </ul> Service configuration secrets are redacted.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
rsalevsky: A few accuracy issues against the actual `role:project:read` grant in aiven-core:

**Drop “View static IP addresses.”** `static_ip_list` is only on the legacy `read_only` / `developer` / `operator` set. This role inherits `project:networking:read`, which is VPCs and peering only.

**These three bullets overstate dedicated API access** (they are only true as fields on project/service GET):
- **View project tags** — `project_tags_list` is not in the role; tags do appear on `project_get`
- **View service tags** — `project_service_tags_list` is not in the role; tags appear on service GET
- **View service users** — usernames appear on service GET with credentials redacted; dedicated user/credential APIs are denied. On this page, “View service users” is mapped to `service:secrets:read` / `service:users:write`, which this role does not have.

I’d fold tags into “view services / project settings” and drop the separate service-users bullet (the secrets sentence already covers credentials). If you keep them, say they are visible on the resource, not via the tags/users APIs.

Nit: move “Service configuration secrets are redacted.” into the list (or a `<br />` note like the Kafka Connect warning on `read_only`). After `</ul>` it is easy to miss in this wide table.

rsalevsky (commented):
Looks like the right row (Console name, API name, sort order, core bundled permissions, and the secrets exclusion all match the product).

Left one inline comment on a couple of overstated capabilities, most importantly static IPs.

docs/platform/concepts/permissions.md
staceysalamon-aiven: I updated the note to make it clear that service user passwords are redacted. As discussed, I'll leave the other permissions as-is (except view static IPs) since this is how we list them elsewhere to show that users can see that field/part of the console.

rsalevsky (approved):

PR #1933, opened 2026-09-01
update: DB service docs landing pages

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-2052


UX-Irene:
Looks good, approved from my side.
Now only Flink is a bit sad in its old structure. But then again a reminder of what used to be.

PR #1958, opened 2026-09-07
update: document S3-compatible object storage support for custom repositories

axeoman:
## Summary

- Note that Aiven for OpenSearch custom repositories support any S3-compatible object storage service (not only Amazon S3), by setting the `endpoint` field.
- Add the `endpoint` field to the S3 custom repository creation example in the Aiven API guide.
- Align the same limitations wording in the native OpenSearch API guide.

## Testing

- Reviewed against the current `custom_repos` S3 settings schema to confirm `endpoint` and its description.
- Manually checked rendered Markdown/MDX for the edited sections.


wojcik-dorota (approved):
Thanks for adding this, @axeoman 💪🏻

PR #1960, opened 2026-09-07
update: cross region support for GCP private connect

wojcik-dorota:
https://aiven.atlassian.net/browse/IP-1242

### DIFF
--- a/docs/platform/howto/use-google-private-service-connect.md
+++ b/docs/platform/howto/use-google-private-service-connect.md
@@ -4,6 +4,7 @@ early: true
 ---
 
 import ConsoleLabel from "@site/src/components/ConsoleIcons";
+import RelatedPages from "@site/src/components/RelatedPages";
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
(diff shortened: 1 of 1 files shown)


docs/platform/howto/use-google-private-service-connect.md
```diff
@@ -227,6 +233,30 @@ Service Connect access for your Aiven for Apache Kafka service while allowing Ai
 Apache Kafka Connect to only be connected via VPC peering.
 :::
 
+## Allow cross-region connections
+
+Private Service Connect endpoints in Google Cloud can accept traffic from other
+regions. You configure this setting, called global access, on your own Private
+Service Connect endpoint.
+
+Keep your Private Service Connect endpoint in the same region as your Aiven
```
heikju, line 242: I found this https://cloud.google.com/blog/products/networking/access-managed-services-globally-with-private-service-connect to be a good and useful explanation of how this works, could be worth including as a related link.

wojcik-dorota: > I found this https://cloud.google.com/blog/products/networking/access-managed-services-globally-with-private-service-connect to be a good and useful explanation of how this works, could be worth including as a related link.

Thanks, @heikju ! Actually, I left it out on purpose: we try to minimize external blog links (opting for official vendor reference docs, which doesn't seem available here). Instead, I've folded the relevant facts directly into the section, so nothing's lost. Hope that makes sense?

docs/platform/howto/use-google-private-service-connect.md
```diff
@@ -227,6 +233,30 @@ Service Connect access for your Aiven for Apache Kafka service while allowing Ai
...
+Compute Engine VMs, Cloud VPN tunnels, or Cloud Interconnect, reach that
+endpoint.
+
+Enable global access on a new or existing endpoint without disrupting traffic.
+To do this, run the following command in the [Google Cloud
+CLI](https://cloud.google.com/sdk/gcloud):
+
+```bash
+gcloud beta compute forwarding-rules update FORWARDING_RULE_NAME \
+  --allow-psc-global-access
```
heikju: According to https://docs.cloud.google.com/sdk/gcloud/reference/compute/forwarding-rules/update this flag seems to be available outside of the `beta` command group too.

heikju (commented):
Seems OK, but the flag seems to be available outside of the `beta` command group to, in case that makes a difference.

docs/platform/howto/use-google-private-service-connect.md
wojcik-dorota: Confirmed, thanks! Checked the GA reference page and `--allow-psc-global-access` is documented there too, no `beta` needed.

PR #1963, opened 2026-09-08
update: default environment variables for Runtime integrations

staceysalamon-aiven:
## Checklist

- [ ] I checked the output.


docs/products/runtime/environment-variables.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Integrated service environment variables
+---
+
+Aiven Runtime provides connection details as environment variables for each connected service.
```
ssuomi-aiven: nit

```suggestion
Aiven Runtime forwards connection details as environment variables for each connected service.
```

or "maps"

docs/products/runtime/environment-variables.md
```diff
@@ -0,0 +1,37 @@
...
+It scans Compose, Containerfile, and Dockerfile manifests during application creation.
```
ssuomi-aiven: Going forward, we should be more clear what is API/MCP functionality and what is Console functionality. "It" is quite ambiguous in this section.

But we already have https://aiven.io/docs/products/runtime/manifest-files/manifests so maybe this should link there too?

docs/products/runtime/environment-variables.md
```diff
@@ -0,0 +1,37 @@
...
+It detects supported data services in Compose files and connects them to your application.
+It reads existing variable names when they include supported connection values. If it
+doesn't find a supported connection value, it suggests conventional variable names based on
+relationships defined in the `depends_on` property of the Compose file.
+
+## Default variables
```
ssuomi-aiven: Let's be clear when we are talking about variables (key + value) or key or value individually.

Also, yes, we are "suggesting these variables" but we are actually _requiring_ them to be defined for each integration.

docs/products/runtime/environment-variables.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Integrated service environment variables
```
ssuomi-aiven: As we already have https://aiven.io/docs/products/runtime/secrets-and-variables as well, which seems to only be about after-create stuff, should we think about the doc structure a bit more?

Currently from the docs, I think the combination of "connected" env vars and custom defined env vars is not that understandable. Maybe it's OK.

staceysalamon-aiven: hmm, I see what you mean, but this is partly why I created a separate section ("Integrations"). To me, this is more of a reference/conceptual doc specific to when the scan is happening. Cross-linking would definitely be a good idea, though, to help users find things via search/chat.

docs/products/runtime/environment-variables.md
staceysalamon-aiven: "It" refers to the subject of the previous sentence, Aiven Runtime. What should it say?

I would need more details on how the interfaces differ and then document them thoroughly.

ssuomi-aiven: The ambiguity is fine when thinking just about Console. But someone integrating with the API directly (directly, over MCP, over CLI) needs to call the scan endpoints manually and then construct a request body to create the service(s) with.

docs/products/runtime/environment-variables.md
```diff
@@ -2,18 +2,19 @@
...
 It reads existing variable names when they include supported connection values. If it
 doesn't find a supported connection value, it suggests conventional variable names based on
 relationships defined in the `depends_on` property of the Compose file.
 
-## Default variables
+## Default environment variables
 
-The following variables are suggested by default. You can customize the variable names.
+The following environment variables are added by default. They're required for the
+integrations, but you can customize the variable names.
```
ssuomi-aiven: This is good 👍

ssuomi-aiven (dismissed):
~Approving because this is already a big positive change.~ **Edit: Found one issue which we should fix**

But we should think about how to make the docs work:
- both for humans and agents
- both for UI use cases and non-UI use cases

Did we have a feature in our docs site already with which you could sprinkle in some machine readable details on the subject?

docs/products/runtime/environment-variables.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Integrated service environment variables
+---
+
+Aiven Runtime maps connection details as environment variables for each connected service.
+It scans Compose, Containerfile, and Dockerfile manifests during application creation.
+It detects supported data services in Compose files and connects them to your application.
+It reads existing variable names when they include supported connection values. If it
+doesn't find a supported connection value, it suggests conventional variable names based on
+relationships defined in the `depends_on` property of the Compose file.
```
ssuomi-aiven: I think this part should be rethought a bit for the same reasons as: https://github.com/aiven/aiven-docs/pull/1963#discussion_r3956659217

ssuomi-aiven: Suggestion:

> Aiven Runtime exposes connection details as environment variables for connected data services. During application creation, you can select a Compose, Containerfile, or Dockerfile manifest to scan. For Compose files, Aiven detects supported data service images and suggests corresponding Aiven services and integrations. If an environment variable in the Compose file points to a detected data service, Aiven uses suggests that variable name, but otherwise a default environment variable name is used.

ssuomi-aiven: Or:

> Aiven Runtime exposes connection details as environment variables for connected data services. During application creation, you can select a Compose, Containerfile, or Dockerfile manifest to scan. For Compose files, Aiven detects supported data service images and suggests corresponding Aiven services and integrations. If an environment variable in the Compose file points to a detected data service, Aiven suggests that variable name. Otherwise, if the service is listed in `depends_on`, a default environment variable name is suggested.

docs/products/runtime/environment-variables.md
staceysalamon-aiven: Isn't that getting into details about what triggers the scan that aren't super relevant here? We definitely need to document these different flows, and need to be clear on the differences for those interfaces in those instructions, but I think getting into that nuance here would add unhelpful complexity. Unless I'm missing something about the differences.

staceysalamon-aiven:
> ~Approving because this is already a big positive change.~ **Edit: Found one issue which we should fix**
> 
> But we should think about how to make the docs work:
> 
> * both for humans and agents
> * both for UI use cases and non-UI use cases
> 
> Did we have a feature in our docs site already with which you could sprinkle in some machine readable details on the subject?

I started investigating ways to do this a couple of weeks ago. I'd like to understand better what is actually going to be discoverable and useful for agents, but the short answer is yes, and I think we should.

docs/products/runtime/environment-variables.md
staceysalamon-aiven: To see what this could look like, I quickly combined the info into one doc: https://github.com/aiven/aiven-docs/pull/1963/changes/2b6685bf9397db02af0d5d1d2d2f088d7b0a28df

I think it might make the most sense to remove the integrations section in that case and link to this doc from the "connect services" instructions. I'm not sure, though.

docs/products/runtime/environment-variables.md
ssuomi-aiven: Well if someone who uses the MCP or API reads this, they would assume "during application creation" (i.e. when POSTing to the `/project/.../service` endpoint) we'd be doing some scanning but actually that happens via different tool calls / endpoints. I'm sure there's some wording we could use that this would be more apparent, e.g. instead of "During application creation" (that's the new copy) we could say "To create an application" 🤷

ssuomi-aiven:
From my limited understanding (it's of course always a bit of a coinflip with these AI overlords) the agents were eager to reach for the docs whenever MCP didn't provide enough guidance. I have no idea how other companies'/projects' docs handle this multi purpose aspect nowadays (I'd ask the AI 🙃 )

ssuomi-aiven (approved):

docs/products/runtime/environment-variables.md
ssuomi-aiven: Hard for me to judge just from the diffs alone. I think it's fine to have a bit of duplicated info but from different angles and entrypoints?

ssuomi-aiven:
I don't have rights to merge here, but if you want to do the one little copy change from https://github.com/aiven/aiven-docs/pull/1963#discussion_r3971546939 then just ping me and I can reapprove.

docs/products/runtime/secrets-and-variables.md
```diff
@@ -19,3 +27,42 @@ you can manually [redeploy your app](/docs/products/runtime/deploy-apps#redeploy
...
+| Aiven for Apache Kafka® | `KAFKA_SECURITY_PROTOCOL` | Set to `SSL`.                      |
+| Aiven for Apache Kafka® | `KAFKA_ACCESS_KEY`        | The access key.                    |
+| Aiven for Apache Kafka® | `KAFKA_ACCESS_CERT`       | The access certificate.            |
+| Aiven for Apache Kafka® | `KAFKA_CA_CERT`           | The trusted CA certificate bundle. |
+
+### Credential formats
+
+Each service type supports one credential format. PostgreSQL, Valkey, and OpenSearch
+provide a complete connection URI. Separate host, port, username, and password
+variables are not supported. Kafka credentials use the five separate variables
```
ssuomi-aiven, line 62:
```suggestion
variables are not supported (coming soon). Kafka credentials use the five separate variables
```

? 😅

ssuomi-aiven (approved):
Ship it ⛵

PR #1976, opened 2026-09-10
fix: reuse and redirects for service-agnostic management section

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-2073


harshini-rangaswamy (approved):
Thanks for this! Looks good to me!

PR #1985, opened 2026-09-11
update: add LibreDB Studio connection guide for PostgreSQL

yusuf-gundogdu:
Adds a connection guide for LibreDB Studio to the PostgreSQL connection methods, alongside the existing DBeaver, DataGrip and pgAdmin pages.

LibreDB Studio is an MIT licensed SQL client you host yourself and open in a browser, so a team connects through one URL instead of installing a client on every machine.

Everything in the guide was tested against a real service on 12 September 2026: a free plan Aiven for PostgreSQL service running Postgres 18.6, with LibreDB Studio 0.15.0 from its Docker image. Pasting the service URI filled in the host, the non standard port, user, password and database name, and set SSL to require because the URI carries sslmode=require. The connection came up in 1139 ms with no workaround, the object browser listed both test tables with correct row counts, a join and group by returned in 482 ms, and EXPLAIN rendered the full plan tree.

Two things in the guide are there because the test surfaced them rather than because they read well. The first run prints the admin password to the container log, which is easy to miss, so the prerequisites say to read it with docker logs. And LibreDB Studio holds a connection pool open, which matters on plans with a low connection limit, so there is a short section pointing at Connections on the service Overview page.

Changes:
- docs/products/postgresql/howto/connect-libredb-studio.md
- sidebars.ts, one entry under Connection methods
- .github/vale/styles/config/vocabularies/Aiven/accept.txt, the product name

Checked before opening: lines wrap at 90, Vale reports no errors, markdownlint is clean, and the commit is signed off.

Happy to add the MySQL page as a separate PR if this one is useful.

Repo: https://github.com/libredb/libredb-studio
Site: https://libredb.org

### DIFF
--- a/.github/vale/styles/config/vocabularies/Aiven/accept.txt
+++ b/.github/vale/styles/config/vocabularies/Aiven/accept.txt
@@ -235,6 +235,7 @@ Klaw
 Kpow
 Kubernetes
 kuromoji
+LibreDB
 LLM
 LLMs
 Loggly
--- a/docs/products/postgresql/howto/connect-libredb-studio.md
+++ b/docs/products/postgresql/howto/connect-libredb-studio.md
@@ -1,0 +1,61 @@
+---
+title: Connect to Aiven for PostgreSQL® with LibreDB Studio
+sidebar_label: LibreDB Studio
+---
+
+import RelatedPages from "@site/src/components/RelatedPages";
+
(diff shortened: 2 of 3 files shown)


wojcik-dorota (approved):
Thanks a ton for adding this, @yusuf-gundogdu 💪🏻 Just made a few tiny style updates, otherwise looks great 👏🏻

